PR #1, opened 2024-09-11
Add Duration and Instant support to Rust.

d3zd3z:
This is a migration of the open PR zephyrproject-rtos/zephyr#78193. To the module version. The only changes from that PR are the squashing a few of the fixup commits.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,4 +56,4 @@ jobs:
         run: |
           cargo --version
 
-          west twister -T samples -v --inline-logs --integration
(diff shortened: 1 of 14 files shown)


d3zd3z:
zephyrproject-rtos/zephyr#78306 incorporates this PR into Zephyr, notably the CI results can be seen there.

ithinuel (approved):

d3zd3z:
This is the same code as zephyrproject-rtos/zephyr#78193, which as an approval from @sigvarthmh, and there is the one from ithinuel so I'll consider this as having two approvals.

PR #3, opened 2024-09-12
Add `cargo doc` generation and publishing.

d3zd3z:


.github/workflows/docs.yml
```diff
@@ -0,0 +1,94 @@
...
+
+jobs:
+  generate-docs:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          path: zephyr-rust-lang
```
pdgendt, line 23: Why not use the module path?

```suggestion
          path: modules/lang/rust
```

d3zd3z: I'm actually not sure it needs any path, since it seems to result in it coming out in zephyr-rust-lang/zephyr-rust-lang. But, now that the module is "real", I wonder if this can just update the module instead of checking it out somewhere else. I'll investigate.

PR #9, opened 2024-10-02
cmake: linking app with runtime library together with rust app library

tejlmand:
The Zephyr PR https://github.com/zephyrproject-rtos/zephyr/pull/78320 clean up and improves link handling in Zephyr.

One improvement is the deterministic and correct link location of runtime and C library linking at the end of link arguments to ensure all libraries are able to use standard functions.

However the librustapp.a library created by the cargo tool also contains runtime functions. It would in most cases be considered correct that a library, such as librustapp.a, which provides functions are linked first so that those functions provided as part of rust toolchain takes precedence over the runtime library provided by the C / C++ toolchain.

However on the riscv32 / riscv64 architecture the rust prebuilt libraries makes use of relocation types in the clzsi2 object which are unknown to the ld linker in use.
Thus linking librustapp.a first, for example like `librustapp.a -lgcc` results in the following warning (and thus CI failures):
```
<path>/ld.bfd: <path>/librustapp.a(45c91108d938afe8-clzdi2.o): unsupported relocation type 0x3d
```

Therefore we ensure that the runtime library is linked first, for example `-lgcc librustapp.a`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -179,7 +179,15 @@ ${config_paths}
         kobj_types_h_target
   )
 
-  target_link_libraries(app PUBLIC -Wl,--allow-multiple-definition ${RUST_LIBRARY})
+  # Linking with the <rt_library> (`$<TARGET_PROPERTY:linker,rt_library>`).
+  # -lgcc / -lcompiler_rt depending on toolchain, linker, and runtime library configuration.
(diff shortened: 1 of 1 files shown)


tejlmand:
see also discussion on discord for further details: https://discord.com/channels/720317445772017664/1290980594359996428

stephanosio (approved):

d3zd3z (approved):

PR #15, opened 2024-10-16
Device Tree Support

d3zd3z:
Built on top of #5, this adds support for the devicetree.

This is definitely a WIP (hence draft), but shows the basics of generating the devicetree.

### DIFF
--- a/samples/blinky/src/lib.rs
+++ b/samples/blinky/src/lib.rs
@@ -11,8 +11,6 @@
 
 use log::warn;
 
-use core::ffi::c_void;
-
 use zephyr::raw::GPIO_OUTPUT_ACTIVE;
 use zephyr::time::{ Duration, sleep };
 
@@ -22,30 +20,6 @@ extern "C" fn rust_main() {
 
     warn!("Starting blinky");
 
-    // Invoke "blink" as a user thread.
-    if false {
-        // Note that for now, this is just a 'false', but is an easy test to use to see if
-        // permissions are correct for usermode Rust.  At this point, the GPIO won't be accessible,
-        // and neither will the memory needed for allocation.
-        unsafe {
-            zephyr::raw::k_thread_user_mode_enter
-                (Some(blink),
-                core::ptr::null_mut(),
-                core::ptr::null_mut(),
-                core::ptr::null_mut());
-        }
-    } else {
-        unsafe {
-            blink(core::ptr::null_mut(),
-                  core::ptr::null_mut(),
-                  core::ptr::null_mut());
-        }
-    }
-}
-
-// fn blink() {
-unsafe extern "C" fn blink(_p1: *mut c_void, _p2: *mut c_void, _p3: *mut c_void) {
-    // Just call a "safe" rust function.
     do_blink();
 }
 


zephyr/src/device.rs
```diff
@@ -0,0 +1,40 @@
...
+/// example, a [`GpioPin`] will reference a single pin, but the underlying device for the gpio
+/// driver will be shared among then.  Generally, the constructor for the individual device will
+/// call `get_instance_raw()` on the underlying device.
+#[allow(dead_code)]
+pub(crate) struct Unique(pub(crate) AtomicUsize);
+
+impl Unique {
+    /// Construct a new unique counter.
+    pub(crate) const fn new() -> Unique {
+        Unique(AtomicUsize::new(0))
```
jilu-ot, line 29: Why not use AtomicBool?

zephyr-build/src/lib.rs
```diff
@@ -73,3 +79,91 @@ pub fn build_kconfig_mod() {
...
+/// print a warning, and then just directly write the file.
+fn write_formatted(file: File, tokens: TokenStream) {
+    let mut rustfmt = Command::new("rustfmt")
+        .args(["--emit", "stdout"])
+        .stdin(Stdio::piped())
+        .stdout(file)
+        .stderr(Stdio::inherit())
+        .spawn()
+        .expect("Failed to run rustfmt");
+    // TODO: Handle the above failing.
```
epilys, line 160: Maybe we can move the `writeln!(out, "{}", tokens).unwrap();` from above here, to handle the error.

zephyr-sys/wrapper.h
```diff
@@ -11,6 +11,12 @@
  * are output.
  */
 
+/*
+ * This is getting build with KERNEL defined, which causes syscalls to not be implemented.  Work
```
epilys, line 15:
```suggestion
 * This is getting built with KERNEL defined, which causes syscalls to not be implemented.  Work
```

zephyr/src/device.rs
```diff
@@ -0,0 +1,40 @@
...
+//! general, these wrap a `*const device` from Zephyr, and provide an API that is appropriate.
+//!
+//! Most of these instances come from the device tree.
+
+use crate::sync::atomic::{AtomicUsize, Ordering};
+
+pub mod gpio;
+pub mod flash;
+
+// Allow dead code, because it isn't required for a given build to have any devices.
```
epilys, line 13: You can make the `allow(dead_code)` attributes to work on the entire module scope by putting `#![allow(dead_code)]` at the top instead of individual definitions/functions

zephyr/src/device/flash.rs
```diff
@@ -0,0 +1,59 @@
...
+        // The `get_instance` on the flash controller would try to guarantee a unique instance,
+        // but in this case, we need one for each device, so just construct it here.
+        // TODO: This is not actually safe.
+        let controller = FlashController { device };
+        Some(FlashPartition { controller, offset, size })
+    }
+}
+
+// Note that currently, the flash partition shares the controller, so the underlying operations
+// are not actually safe.  Need to rethink how to manage this.
```
epilys, line 59: Maybe move this comment on the top of the file?

zephyr/src/device/gpio.rs
```diff
@@ -0,0 +1,119 @@
+//! Most devices in Zephyr operate on a `struct device`.  This provides untyped access to
+//! devices.  We want to have stronger typing in the Zephyr interfaces, so most of these types
+//! will be wrapped in another structure.  This wraps a Gpio device, and provides methods to
+//! most of the operations on gpios.
+//!
+//! Safey: In general, even just using gpio pins is unsafe in Zephyr.  The gpio drivers are used
```
epilys, line 6:
```suggestion
//! # Safety
//! 
//! In general, even just using gpio pins is unsafe in Zephyr.  The gpio drivers are used
```

zephyr-build/src/lib.rs
```diff
@@ -73,3 +79,91 @@ pub fn build_kconfig_mod() {
...
+    let generated = format!("{}/devicetree_generated.h", gen_include);
+    DeviceTree::new(&zephyr_dts, generated)
+}
+
+pub fn build_dts() {
+    let dt = import_dt();
+
+    let outdir = env::var("OUT_DIR").expect("OUT_DIR must be set");
+    let out_path = Path::new(&outdir).join("devicetree.rs");
+    let mut out = File::create(&out_path).expect("Unable to create devicetree.rs");
```
epilys, line 98: You can write a comment at the top of the file to say it's generated (even if we try to run rustfmt afterwards; but we should pass `format_generated_files=true` to rustfmt in that case) this can tell format linters in general that there should be no style/formatting checks in this file.

This will help if the file was generated without rustfmt being present, say in a CI environment.

```rust
// @generated
```

For more information see <https://github.com/rust-lang/rustfmt/pull/4296>

samples/blinky/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2024 Linaro LTD
+# SPDX-License-Identifier: Apache-2.0
+
+[package]
+# This must be rustapp for now.
+name = "rustapp"
+version = "0.1.0"
+edition = "2021"
+description = "A sample hello world application in Rust"
+license = "Apache-2.0 or MIT"
```
epilys, line 10:
```suggestion
license = "Apache-2.0 OR MIT"
```

See <https://spdx.github.io/spdx-spec/v2-draft/SPDX-license-expressions/#d2-case-sensitivity>

samples/blinky/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2024 Linaro LTD
+# SPDX-License-Identifier: Apache-2.0
+
+[package]
+# This must be rustapp for now.
+name = "rustapp"
+version = "0.1.0"
+edition = "2021"
+description = "A sample hello world application in Rust"
```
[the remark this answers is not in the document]
epilys: Probably "Blink a LED forever using the GPIO API." like mentioned in the `README.rst` file?

samples/blinky/README.rst
```diff
@@ -0,0 +1,97 @@
...
+   		led0 = &myled0;
+   	};
+
+   	leds {
+   		compatible = "gpio-leds";
+   		myled0: led_0 {
+   			gpios = <&gpio0 13 GPIO_ACTIVE_LOW>;
+                };
+   	};
+   };
```
epilys, line 78: There is mixed tab and spaces indentation here, I assume by accident

samples/blinky/src/lib.rs
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) 2024 Linaro LTD
+// SPDX-License-Identifier: Apache-2.0
+
+#![no_std]
+
+// Sigh. The check config system requires that the compiler be told what possible config values
+// there might be.  This is completely impossible with both Kconfig and the DT configs, since the
+// whole point is that we likely need to check for configs that aren't otherwise present in the
+// build.  So, this is just always necessary.
+#![allow(unexpected_cfgs)]
```
epilys, line 10: You can put the expected `cfg` and `cfg` values in `[lints.*]` in `Cargo.toml`:

https://blog.rust-lang.org/2024/05/06/check-cfg.html

samples/blinky/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+    } else {
+        unsafe {
+            blink(core::ptr::null_mut(),
+                  core::ptr::null_mut(),
+                  core::ptr::null_mut());
+        }
+    }
+}
+
+// fn blink() {
```
epilys, line 44:
```suggestion
```

samples/blinky/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+        unsafe { led0.toggle_pin(&mut gpio_token); }
+        sleep(duration);
+    }
+}
+
+#[cfg(not(dt = "aliases::led0"))]
+fn do_blink() {
+    warn!("No leds configured");
+    loop {
+    }
```
epilys, line 76:
```suggestion
    loop {}
```

samples/blinky/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+fn do_blink() {
+    warn!("Inside of blinky");
+
+    let mut led0 = zephyr::devicetree::aliases::led0::get_instance().unwrap();
+    let mut gpio_token = unsafe { zephyr::device::gpio::GpioToken::get_instance().unwrap() };
+
+    if !led0.is_ready() {
+        warn!("LED is not ready");
+        loop {
+        }
```
epilys, line 60:
```suggestion
        loop {}
```

samples/blinky/src/main.c
epilys, line 1: Why is this file included? As a comparison with C? It's not mentioned in the README

zephyr-build/src/devicetree.rs
```diff
@@ -0,0 +1,306 @@
...
+mod output;
+mod parse;
+
+pub use augment::{Augment, load_augments};
+
+pub struct DeviceTree {
+    /// The root of the tree.
+    root: Rc<Node>,
+    /// All of the labels.
+    labels: BTreeMap<String, Rc<Node>>,
```
epilys, line 37: Any reason why this is not a `HashMap`? Do we want it to be sorted? <https://doc.rust-lang.org/std/collections/index.html#use-a-btreemap-when>

zephyr-build/src/devicetree.rs
```diff
@@ -0,0 +1,306 @@
...
+    pub value: Vec<Value>,
+}
+
+// Although the real device flattends all of these into bytes, Zephyr takes advantage of them at a
+// slightly higher level.
+#[derive(Debug)]
+pub enum Value {
+    Words(Vec<Word>),
+    Bytes(Vec<u8>),
+    Phandle(Phandle), // TODO
```
epilys, line 72: Leftover `TODO`?

samples/blinky/CMakeLists.txt
epilys, line 1: The file contains an `Apache-2.0` SPDX header but the code is `Apache-2.0 OR MIT` according to the cargo manifest

zephyr-build/src/devicetree.rs
```diff
@@ -0,0 +1,306 @@
...
+
+pub struct DeviceTree {
+    /// The root of the tree.
+    root: Rc<Node>,
+    /// All of the labels.
+    labels: BTreeMap<String, Rc<Node>>,
+}
+
+// This is a single node in the devicetree.
+pub struct Node {
```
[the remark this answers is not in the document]
d3zd3z: I'm not too worried, as all of this code runs in the host once at build time. I'd probably only be concerned about that for String on the target.

zephyr-build/src/devicetree.rs
```diff
@@ -0,0 +1,306 @@
...
+    /// Decode the zephyr.dts and devicetree_generated.h files from the build and build an internal
+    /// representation of the devicetree itself.
+    pub fn new<P1: AsRef<Path>, P2: AsRef<Path>>(dts_path: P1, dt_gen: P2) -> DeviceTree {
+        let ords = OrdMap::new(dt_gen);
+
+        let dts = std::fs::read_to_string(dts_path)
+            .expect("Reading zephyr.dts file");
+        let dt = parse::parse(&dts, &ords);
+        dt.resolve_phandles();
+        dt.set_parents();
```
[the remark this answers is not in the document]
d3zd3z: That probably makes sense. When I started with them, I wasn't sure how large they would end up.

samples/blinky/src/lib.rs
d3zd3z: This should get fixed when I do a rustfmt cleanup.

samples/blinky/src/main.c
d3zd3z: This is an error. I'll remove it.

zephyr-build/src/devicetree.rs
d3zd3z: I'll check.

samples/blinky/CMakeLists.txt
d3zd3z: Good catch. I think what I'm going for is to make the Rust contributions Apache-2.0 OR MIT, and the Zephyr contributions just Apache-2.0. When I first wrote it, this was in the main zephyr tree, so it was less clear.

zephyr-build/src/devicetree.rs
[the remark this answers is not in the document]
epilys: Agreed, it was just a nitpick :)

zephyr-build/src/devicetree.rs
epilys: That's a good reason. Maybe you can add a comment to mention that `BTreeMap` keeps its values sorted, so the output in the generated files is always deterministic.

samples/blinky/sample.yaml
```diff
@@ -0,0 +1,12 @@
+sample:
```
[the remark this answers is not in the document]
d3zd3z: Hmm, I've not seen that done in any other sample. But, I'll add the path to the docs that explains it.

zephyr/src/device.rs
d3zd3z: Done. I wanted to verify that 'false' is guaranteed to be a zero value in rust, which it is. There are some initializations (esp of statics) where we're not actually using an initializer, but zero-initted memory.

zephyr-build/src/devicetree.rs
```diff
@@ -0,0 +1,306 @@
...
+use std::{cell::RefCell, collections::BTreeMap, path::Path, rc::Rc};
+
+mod augment;
+mod ordmap;
+mod output;
+mod parse;
+
+pub use augment::{Augment, load_augments};
+
+pub struct DeviceTree {
```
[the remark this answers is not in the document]
d3zd3z: I'll add, I'm confused by the weird way github's diff shows it.

zephyr-build/src/devicetree.rs
```diff
@@ -0,0 +1,306 @@
...
+    fn phandle_walk(&self, labels: &BTreeMap<String, Rc<Node>>) {
+        for prop in &self.properties {
+            for value in &prop.value {
+                value.phandle_walk(labels);
+            }
+        }
+        for child in &self.children {
+            child.phandle_walk(labels);
+        }
+    }
```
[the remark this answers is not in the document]
d3zd3z: This runs at build time on the host. There is no particular reason to avoid recursion.

zephyr-build/src/devicetree.rs
```diff
@@ -0,0 +1,306 @@
...
+
+    /// Get this property in its entirety.
+    fn get_property(&self, name: &str) -> Option<&[Value]> {
+        for p in &self.properties {
+            if p.name == name {
+                return Some(&p.value);
+            }
+        }
+        return None;
```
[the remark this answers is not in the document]
d3zd3z: Almost works. Needs `Some(p.as_slice())` to actually get the slice instead of a reference to a Vec.

zephyr-build/src/devicetree.rs
```diff
@@ -0,0 +1,306 @@
...
+            .and_then(|p| {
+                match p {
+                    &[Value::Words(ref w)] => Some(w.as_ref()),
+                    _ => None,
+                }
+            })
+    }
+
+    /// Get a property that consists of a single number.
+    fn get_number(&self, name: &str) -> Option<u32> {
```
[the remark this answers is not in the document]
d3zd3z: Not sure this is exactly a getter, though, as it takes an argument.

zephyr-build/src/devicetree.rs
```diff
@@ -0,0 +1,306 @@
...
+            Value::Phandle(ph) => ph.phandle_resolve(labels),
+            Value::Words(words) => {
+                for w in words {
+                    match w {
+                        Word::Phandle(ph) => ph.phandle_resolve(labels),
+                        _ => (),
+                    }
+                }
+            }
+            _ => (),
```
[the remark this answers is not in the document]
d3zd3z: Really? I've never seen that.

zephyr-build/src/devicetree/output.rs
```diff
@@ -0,0 +1,215 @@
...
+        }
+        general_property(prop)
+    }
+}
+
+impl Node {
+    /// Return the route to this node, as a Rust token stream giving a fully resolved name of the
+    /// route.
+    pub fn route_to_rust(&self) -> TokenStream {
+        let route: Vec<_> = self.route.iter().map(|p| dt_to_lower_id(p)).collect();
```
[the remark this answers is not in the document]
d3zd3z: This change fails to compile, quite violently at that.

zephyr-build/src/lib.rs
d3zd3z: I'll think about how to best do this. Really, this should all change to be a real proc macro instead of outputting to a file that gets included.

zephyr/src/device.rs
d3zd3z: I guess in this case, everything is possibly dead, except for 'new'. I suppose it makes sense to allow for a build with no devices.

samples/blinky/src/lib.rs
d3zd3z: This isn't possible, as we don't know the set of possible cfgs. This is actually kind of something broken about the way Rust is doing it. Other than scanning the entire Zephyr Kconfig tree for possible configs, we have no way of knowing what things the user's code _might_ want to be able to config on.

Now, admittedly an app could add their own, and maybe we should do that? Is that what you are suggesting?

samples/blinky/src/lib.rs
d3zd3z: I'll let this get fixed when I do the big rustfmt/clippy fixes after this merges.

samples/blinky/README.rst
theotherjimmy: github's render looks to have correct spacing.

samples/blinky/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+        // return;
```
theotherjimmy, line 60:
```suggestion
```

samples/blinky/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+use zephyr::time::{ Duration, sleep };
+
+#[no_mangle]
+extern "C" fn rust_main() {
+    unsafe { zephyr::set_logger().unwrap(); }
+
+    warn!("Starting blinky");
+
+    // Invoke "blink" as a user thread.
+    if false {
```
theotherjimmy, line 26: uh, this never happens. Is there a plan make this code reachable?

zephyr-build/src/devicetree.rs
```diff
@@ -0,0 +1,291 @@
...
+    fn compatible_path(&self, path: &[Option<&str>]) -> bool {
+        self.path_walk(path, 0)
+    }
+
+    /// Recursive path walk, to make borrowing simpler.
+    fn path_walk(&self, path: &[Option<&str>], pos: usize) -> bool {
+        if pos >= path.len() {
+            // Once past the end, we consider everything a match.
+            return true;
+        }
```
theotherjimmy, line 152:
```suggestion
        // It's a programming error to continue searching after finding the target.
        assert!(pos <= path.len());
        if pos == path.len() {
            return true;
        }
```
This is a tad pedantic, and I think this is what's actually meant.

zephyr-build/src/devicetree.rs
```diff
@@ -0,0 +1,291 @@
...
+        // Walk down the tree.  We have to check for None here, as we can't recurse on the none
+        // case.
+        if let Some(child) = self.parent.borrow().as_ref() {
+            child.path_walk(path, pos + 1)
+        } else {
+            // We've run out of nodes, so this is considered not matching.
+            false
+        }
+    }
+
```
theotherjimmy, line 168:
```suggestion
    fn path_walk(&self, path: &[Option<&str>]) -> bool {
        match path.first() {
            None => return true;
            Some(current) if matches!(current, Some(name) if !self.is_compatible(name)) => return false;
            _ => ();
        };

        // Walk down the tree.  We have to check for None here, as we can't recurse on the none
        // case.
        if let Some(child) = self.parent.borrow().as_ref() {
            child.path_walk(path[1:])
        } else {
            // We've run out of nodes, so this is considered not matching.
            false
        }
    }

```

Alternately, use slicing to avoid storing and worrying about the position.

samples/blinky/README.rst
```diff
@@ -0,0 +1,97 @@
...
+
+You will see a build error at the source code line defining the ``struct
+gpio_dt_spec led`` variable if you try to build Blinky for an unsupported
+board.
+
+On GCC-based toolchains, the error looks like this:
+
+.. code-block:: none
+
+   error: '__device_dts_ord_DT_N_ALIAS_led_P_gpios_IDX_0_PH_ORD' undeclared here (not in a function)
```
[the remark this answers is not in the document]
d3zd3z: Well, the linker specifically. That's the lovely names in the generated C code that we reference in the `get_instance`.

samples/blinky/src/lib.rs
d3zd3z: I can probably just remove it. At one point I was trying to make more of this work with userspace. At this point, since userspace is incompatible with allocated kernel objects in Zephyr, I'm kind of just not supporting it at this time. I can just remove it.

zephyr-build/src/devicetree.rs
d3zd3z: I've rewritten the `compatible_path` to just call itself with the slice instead of needing a helper.

PR #22, opened 2024-10-18
A perpetual WIP PR

d3zd3z:
This PR holds onto changes that would be considered "work in progress", but can be reviewed.  Ideally, this should be based on top of other pending PRs so that the actual change here is kept minimal.

Built upon #15


d3zd3z:
This is quite old.

PR #28, opened 2024-11-05
cmake: Avoid appearing to support targets with dynamic timers

d3zd3z:
The code conditionally compiles against building on systems with dynami timers.  Until this support can be added, be sure to not define `RUST_SUPPORTED` on these targets.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,9 +9,11 @@ on:
   push:
     branches:
       - main
+      - v*-branch
   pull_request:
     branches:
       - main
+      - v*-branch
   schedule:
     - cron: "0 0 * * *"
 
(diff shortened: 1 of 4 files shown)


teburd (approved):
I believe this is fixing what could be considered a bug in CI today, if so is this worth tagging bug and getting in for 4.0?

mmahadevan108 (approved):

ycsin (approved):

d3zd3z:
> I believe this is fixing what could be considered a bug in CI today, if so is this worth tagging bug and getting in for 4.0?

I think we do need to come up with a mechanism to tell CI what platforms actually make sense to test. There is some discussion about tagging to do this. We could also make the `CONFIG_RUST_SUPPORTED` a specific list of supported boards instead of just all of the targets. Thoughts?

This particular change does solve a Kconfig compatibility issue with that.

There are a few more problems:
- There are some boards that need toolchains I didn't get included in the Docker image
- A few targets have some weird errors in bindgen.

cfriedt (approved):

PR #33, opened 2024-11-26
Implement Bounded channels

### LINKED ISSUE #34: sync::channel, bounded

d3zd3z:
Whereas the existing unbounded channels perform an allocation for each message sent, these bounded channels do a single allocation on initialization (of the channel, the zephyr queues used to implement it do an allocation). This allows send to block as well as recv.

This is the first part, coming soon will be send and receive with a timeout. This will allow using bounded channels from irq context (with no timeout).

Fixes #34

### DIFF
--- a/zephyr/src/sync/channel.rs
+++ b/zephyr/src/sync/channel.rs
@@ -22,6 +22,21 @@
 //! The requirement for Drop is only strictly true if the IRQ handler calls `try_recv` and drops
 //! received message.  If the message is *always* sent over another channel or otherwise not
 //! dropped, it *might* be safe to use these messages.
+//!
+//! ## Dropping of Sender/Receiver
+//!
+//! Crossbeam channels support detecting when all senders or all receivers have been dropped on a
+//! channel, which will cause the handles on the other end to error, including waking up current
+//! threads waiting on those channels.
+//!
+//! At this time, this isn't implementable in Zephyr, as there is no API to wake up all threads
+//! blocked on a given `k_queue`.  As such, this scenario is not supported.  What actually happens
+//! is that when all senders or receivers on a channel are dropped, operations on the other end of
+//! the channel may just block (or queue forever with unbounded queues).  If all handles (both
+//! sender and receiver) are dropped, the last drop will cause a panic.  It maybe be better to just
+//! leak the entire channel, as any data associated with the channels would be leaked at this point,
+//! including the underlying Zephyr `k_queue`.  Until APIs are added to Zephyr to allow the channel
+//! information to be safely freed, these can't actually be freed.
 
 extern crate alloc;
 
@@ -187,15 +202,14 @@ impl<T> Drop for Sender<T> {
             SenderFlavor::Unbounded { queue, .. } => {
                 unsafe {
                     queue.release(|_| {
-                        crate::printkln!("Release");
-                        true
+                        panic!("Unbounded queues cannot currently be dropped");
                     })
                 }
             }
             SenderFlavor::Bounded(chan) => {
                 unsafe {
                     chan.release(|_| {
-                        panic!("Bounded queues cannot be dropped");
+                        panic!("Bounded queues cannot currently be dropped");
                     })
                 }
             }
--- a/zephyr/src/sync/channel.rs
+++ b/zephyr/src/sync/channel.rs
@@ -103,7 +103,7 @@ pub fn unbounded<T>() -> (Sender<T>, Receiver<T>) {
 /// with a capacity of zero.
 pub fn bounded<T>(cap: usize) -> (Sender<T>, Receiver<T>) {
     if cap == 0 {
-        panic!("Zero capacity queues no supported on Zephyr");
+        panic!("Zero capacity queues are not supported on Zephyr");
     }
 
     let (s, r) = counter::new(Bounded::new(cap));
--- a/zephyr/src/sync/channel.rs
+++ b/zephyr/src/sync/channel.rs
@@ -10,6 +10,18 @@
 //! In other words, `zephyr::sys::Queue` is a Rust friendly implementation of `k_queue` in Zephyr.
 //! This module provides `Sender` and `Receiver`, which can be cloned and behave as if they had an
 //! internal `Arc` inside them, but without the overhead of an actual Arc.
+//!
+//! ## IRQ safety
+//!
+//! These channels are usable from IRQ context on Zephyr in very limited situations.  Notably, all
+//! of the following must be true:
+//! - The channel has been created with `bounded()`, which pre-allocates all of the messages.
+//! - If the type `T` has a Drop implementation, this implementation can be called from IRQ context.
+//! - Only `try_send` or `try_recv` are used on the channel.
+//!
+//! The requirement for Drop is only strictly true if the IRQ handler calls `try_recv` and drops
+//! received message.  If the message is *always* sent over another channel or otherwise not
+//! dropped, it *might* be safe to use these messages.
 
 extern crate alloc;
 
--- a/zephyr/src/sync/channel.rs
+++ b/zephyr/src/sync/channel.rs
@@ -22,7 +22,7 @@ use core::marker::PhantomData;
 use core::mem::MaybeUninit;
 
 use crate::sys::queue::Queue;
-use crate::time::Forever;
+use crate::time::{Forever, NoWait, Timeout};
 
 mod counter;
 
@@ -118,9 +118,15 @@ unsafe impl<T: Send> Send for Sender<T> {}
 unsafe impl<T: Send> Sync for Sender<T> {}
 
 impl<T> Sender<T> {
-    /// Sends a message over the given channel. This will perform an alloc of the message, which
-    /// will have an accompanied free on the recipient side.
-    pub fn send(&self, msg: T) -> Result<(), SendError<T>> {
+    /// Waits for a message to be sent into the channel, but only for a limited time.
+    ///
+    /// This call will block until the send operation can proceed or the operation times out.
+    ///
+    /// For unbounded channels, this will perform an allocation (and always send immediately).  For
+    /// bounded channels, no allocation will be performed.
+    pub fn send_timeout<D>(&self, msg: T, timeout: D) -> Result<(), SendError<T>>
+        where D: Into<Timeout>,
+    {
         match &self.flavor {
             SenderFlavor::Unbounded { queue, .. } => {
                 let msg = Box::new(Message::new(msg));
@@ -131,7 +137,10 @@ impl<T> Sender<T> {
             }
             SenderFlavor::Bounded(chan) => {
                 // Retrieve a message buffer from the free list.
-                let buf = unsafe { chan.free.recv(Forever) };
+                let buf = unsafe { chan.free.recv(timeout) };
+                if buf.is_null() {
+                    return Err(SendError(msg));
+                }
                 let buf = buf as *mut Message<T>;
                 unsafe {
                     buf.write(Message::new(msg));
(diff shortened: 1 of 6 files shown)


zephyr/src/sync/channel.rs
```diff
@@ -144,38 +169,61 @@ impl<T> Receiver<T> {
...
-        unsafe {
-            self.queue.release(|_| {
-                crate::printkln!("Release");
-                true
-            })
+        match &self.flavor {
+            ReceiverFlavor::Unbounded { queue, .. } => {
+                unsafe {
+                    queue.release(|_| {
+                        crate::printkln!("Release");
```
Ablu, line 315: This reads like it is just a temporary debug print?

zephyr/src/sync/channel.rs
```diff
@@ -92,45 +95,67 @@ impl<T> Sender<T> {
...
+}
+
+/// The "flavor" of a sender.  This maps to the type of channel.
+enum SenderFlavor<T> {
+    /// An unbounded queue.  Messages are allocated with Box, and sent directly.
+    Unbounded {
+        queue: counter::Sender<Queue>,
+        _phantom: PhantomData<T>,
     }
 }
```
Ablu, line 232: nit: I personally would have put them over the type where they were used (same below, would have kept Sender + Receiver Flavor together).

zephyr/src/sync/channel.rs
```diff
@@ -224,6 +290,45 @@ enum ReceiverFlavor<T> {
...
+            .map(|_| {
+                UnsafeCell::new(MaybeUninit::uninit())
+            })
+        .collect();
+
+        let free = Queue::new().unwrap();
+        let chan = Queue::new().unwrap();
+
+        // Add each of the boxes to the free list.
+        for chan in &slots {
```
Ablu, line 393: why `chan`? Aren't these all part of a single channel? I find it a bit confusing as it is hiding the previous `chan`.

zephyr/src/sync/channel.rs
```diff
@@ -224,6 +290,45 @@ enum ReceiverFlavor<T> {
...
+            unsafe {
+                free.send(chan.get() as *mut c_void);
```
Ablu, line 395: I find this pretty tough to review without SAFETY comments. The Box itself does not disallow moving the contents of it (through swap). Yet, this relies on pointers into it. So should the Slot or Box be wrapped in `Pin`?

I know that it does not change anything in practice. But I feel like it makes it a bit easier to reason about.

zephyr/src/sys/queue.rs
```diff
@@ -62,8 +62,14 @@ impl Queue {
     /// Get an element from a queue.
     ///
     /// This routine removes the first data item from the [`Queue`].
-    pub unsafe fn recv(&self) -> *mut c_void {
-        k_queue_get(self.item.get(), K_FOREVER)
+    /// The timeout value can be [`Forever`] to block until there is a message, [`NoWait`] to check
+    /// and immediately return if there is no message, or a [`Duration`] to indicate a specific
+    /// timeout.
+    pub unsafe fn recv<T>(&self, timeout: T) -> *mut c_void
```
Ablu, line 68: not relevant for this comment, but: `pub unsafe`? Should this be `pub(crate) unsafe` (probably the same for the other functions on this impl)? It reads like it is just an internal wrapper.

Ablu (approved):
Only some nitpicking comments. LGTM overall. But I will repeat myself by saying that I find the safety tough to review without // SAFETY comments :).

zephyr/src/sync/channel.rs
d3zd3z: Honestly, maybe it should just be a panic. Not really sure what to do other than just leak the underlying channel. In order to actually close them, Zephyr needs a new API to be able to wake up _all_ blockers on a queue, not just one of them.

zephyr/src/sync/channel.rs
d3zd3z: Perhaps I should move them to their own file, that's how crossbeam does it. But crossbeam also has many more flavors, and the code is a lot more complex, as they are actually implementing the channel, not just using an underlying `k_queue` like this does.

zephyr/src/sync/channel.rs
d3zd3z: I wasn't quite sure what to call it. I think I'll just change it to `for slot in &slots`, which should help.

zephyr/src/sync/channel.rs
d3zd3z: It could be, which might make it clearer. But, it isn't public, so doesn't leak anything. I'll add the pin.

zephyr/src/sys/queue.rs
d3zd3z: No, definitely 'pub'. This is part of the public interface. The `k_queue` is perfectly usable outside of this crate, it is just unsafe to do so. But there is no reason to stop external users from being able to do so.

zephyr/src/sync/channel.rs
d3zd3z: Actually, I'm not quite sure I understand where you would put this definition. You would group the flavors together rather than keeping the sender and receiver things together. I'm not sure I'd find that clearer.

d3zd3z:
I have updated based on the review feedback, other than one that I'm not sure I agree with. These changes cosmetic, mostly adding comments. The `Pin` of the bounded message channel has some code change, but has no effect on behavior.

zephyr/src/sync/channel.rs
```diff
@@ -58,6 +93,29 @@ pub fn unbounded<T>() -> (Sender<T>, Receiver<T>) {
...
+///
+/// The messages are allocated at channel creation time.  If there are no messages at `send` time,
+/// send will block (possibly waiting for a timeout).
+///
+/// At this time, Zephyr does not support crossbeam's 0 capacity queues, which are also called
+/// a rendezvous, where both threads wait until in the same region.  `bounded` will panic if called
+/// with a capacity of zero.
+pub fn bounded<T>(cap: usize) -> (Sender<T>, Receiver<T>) {
+    if cap == 0 {
+        panic!("Zero capacity queues no supported on Zephyr");
```
jilu-ot, line 106: Maybe change message to: Zero capacity queues are not supported on Zephyr

zephyr/src/sys/queue.rs
jilu-ot: Not sure, but if the intention is to create a safe wrapper for unsafe Zephyr API calls.

Would it not be better to wrap the k_queue_append call in an unsafe block, instead of declaring the send unsafe?

Otherwise the user of this function will have to wrap the code using this API in an unsafe block.

The same goes for the other unsafe APIs.

zephyr/src/sync/channel.rs
```diff
@@ -81,101 +139,320 @@ impl<T> Message<T> {
...
+                SenderFlavor::Bounded(chan.acquire())
+            }
+        };
+
+        Sender { flavor }
     }
 }
 
+/// The "flavor" of a sender.  This maps to the type of channel.
+enum SenderFlavor<T> {
```
jilu-ot, line 252: I think it would be more clear to a user what this type means, if it was called something like ChannelType.

zephyr/src/sync/channel.rs
```diff
@@ -81,101 +139,320 @@ impl<T> Message<T> {
...
 unsafe impl<T: Send> Send for Receiver<T> {}
 unsafe impl<T: Send> Sync for Receiver<T> {}
 
 impl<T> Receiver<T> {
+    /// Waits for a message to be received from the channel, but only for a limited time.
+    ///
+    /// If the channel is empty and not disconnected, this call will block until the receive
+    /// operation can proceed or the operation times out.
+    /// wake up and return an error.
+    pub fn recv_timeout<D>(&self, timeout: D) -> Result<T, RecvError>
```
jilu-ot, line 283: How about calling this method receive and let the timeout be an Option?

Makes the name more easily read and also covers the recv function.

zephyr/src/sync/channel.rs
```diff
@@ -81,101 +139,320 @@ impl<T> Message<T> {
...
+    }
+
+    /// Attempts to receive a message from the channel without blocking.
+    ///
+    /// This method will either receive a message from the channel immediately, or return an error
+    /// if the channel is empty.
+    ///
+    /// This method is safe to use from IRQ context, if and only if the channel was created as a
+    /// bounded channel.
+    pub fn try_recv(&self) -> Result<T, RecvError> {
```
jilu-ot, line 339: This method could be called get_message to more clearly say that it is retrieving a message from the channel, if there is one.

zephyr/src/sync/channel.rs
```diff
@@ -81,101 +139,320 @@ impl<T> Message<T> {
...
+impl<T> Bounded<T> {
+    fn new(cap: usize) -> Self {
+        let slots: Box<[Slot<T>]> = (0..cap)
+            .map(|_| {
+                UnsafeCell::new(MaybeUninit::uninit())
+            })
+        .collect();
+        let slots = Box::into_pin(slots);
+
+        let free = Queue::new().unwrap();
```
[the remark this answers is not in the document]
d3zd3z: At least in stable rust, allocation failures result in a panic. There is nightly provision for handling it, but it doesn't seem to be through returns.

zephyr/src/sync/channel.rs
d3zd3z: Perhaps, but I'm trying to keep the naming and such as close to the crossbeam-channel implementation as possible, where it is called this.

zephyr/src/sync/channel.rs
d3zd3z: Again, I'm trying to be compatible with crossbeam-channel, which has defined the API this way. The timeouts in Zephyr already support NoWait and Forever, but doing it this way allows the code to be used on the host with crossbeam-channel without changes.

zephyr/src/sync/channel.rs
d3zd3z: Likewise, just keeping compatible with an existing implementation.

zephyr/src/sync/channel.rs
jilu-ot: Ok.

zephyr/src/sync/channel.rs
jilu-ot: Ok.

zephyr/src/sync/channel.rs
jilu-ot: I guess it would be even better with: Zero capacity queues are not supported by Zephyr

cfriedt (approved):

zephyr/src/sys/queue.rs
d3zd3z: The problem is that `k_queue_append` is _never_ safe. It takes a lot of stuff around it to make it safe. The idea is that the `sys` types are rust-language interfaces, but are still unsafe. Then in `sync` or other places will be the usable interfaces. The `sys` ones are available as there might be cases where someone needs the lower level functionality.

zephyr/src/sync/channel.rs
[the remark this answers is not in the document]
d3zd3z: > Ok. If it is not the convention. I just thought it might be better to let the user of these modules decide what should happen in this case.

I think we'll have to wait for Rust development itself to figure out how they're going to solve this. Nothing in 'alloc' returns an error on memory exhaustion. They call to a function that can either panic overall, or stop the present thread.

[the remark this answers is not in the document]
jilu-ot: Ok.

jilu-ot (approved):

PR #36, opened 2024-12-04
Add bt_ APIs to the generated bindings

jilu-ot:
Add bt_ APIs to the generated bindings to enable bluetooth application development.


d3zd3z (approved):

Ablu (approved):
Would probably have added things in alphabetical order. But it looks like that was not a  convention before this anyway :).

PR #38, opened 2024-12-09
cmake: use project name instead of hardcoded name

inthehack:
This commit offers to give an arbitrary name to the Rust application project under CMake.

Indeed, before this commit, the CMake project for the Rust application had the name `librustapp`. Now, it can be set with any value.

### DIFF
--- a/samples/hello_world/sample.yaml
+++ b/samples/hello_world/sample.yaml
@@ -1,6 +1,6 @@
 sample:
   description: Hello world, but in Rust
-  name: hello rust world
+  name: hello world
 common:
   harness: console
   harness_config:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,8 +86,9 @@ function(rust_cargo_application)
   endif()
   set(BUILD_LIB_DIR "${CMAKE_CURRENT_SOURCE_DIR}/${RUST_TARGET}/${RUST_BUILD_TYPE}")
 
+  set(RUST_LIBRARY_NAME "${CMAKE_PROJECT_NAME}")
   set(CARGO_TARGET_DIR "${CMAKE_CURRENT_BINARY_DIR}/rust/target")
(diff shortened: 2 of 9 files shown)


d3zd3z (approved):

d3zd3z:
So, I think this is a good idea, but it is causing every single CI test to fail. I think you'll also need to fix the app name in every Cargo.toml file in the tree.

samples/hello_world/sample.yaml
```diff
@@ -1,6 +1,6 @@
 sample:
   description: Hello world, but in Rust
-  name: hello rust world
+  name: hello world
```
d3zd3z, line 3: I'll be real curious about this. At one point, I ran into twister failures if there were two samples, across the whole system, that had the same name.

d3zd3z: So, yes, this is real. Twister uses the name field as a key, and if two samples have the same name, across the whole tree, some operations may silently ignore one of them.

MFarabi619:
Hello hello, any movement of this? Would be happy to jump in to help push this along.

inthehack:
> Hello hello, any movement of this? Would be happy to jump in to help push this along.

@MFarabi619 Hi, yes feel free to update this PR and resolve conflicts. Sadly, I do not have time to work on it right now. You can ask any question if you need help.

MFarabi619:
Sounds good, feel free to assign the issue (if there is one) to me. I've got some patches locally addressing this with a similar approach.

Did you want me to: 

- make a new branch from trunk, or
- stack on this, or
- continue where this one left off via rebase and push access?

inthehack:
From my point of view, go the way that is easier and faster for you. If you create a new PR, just reference it here to help people follow up.

d3zd3z:
> * make a new branch from trunk, or

For the most part, you'll probably find it easier to make a new PR with your changes. You can cherry pick @inthehack changes. And, just post a reference to the new PR here.

felipebalbi:
Last update here was April 29. @MFarabi619 do you still want to work on this. @d3zd3z, sounds like this could be closed, maybe?

PR #39, opened 2024-12-09
Feat add custom export of c bindings

jpeeters:


jpeeters:
Hi @d3zd3z 👋🏼 !

Sorry to ping you. This is my first PR and I don't know who to ask for review. Do you have time to check it out?

zephyr-build/src/lib.rs
```diff
@@ -12,30 +12,47 @@
...
+use std::collections::HashSet;
 use std::env;
 use std::fs::File;
 use std::path::Path;
 
 use regex::Regex;
 
+/// Extract boolean Kconfig entries.  This must happen in any crate that wishes to access the
+/// configuration settings.
+pub fn extract_kconfig_bool_options(path: &str) -> anyhow::Result<HashSet<String>> {
```
d3zd3z, line 24: I notice you have also changed other Result to anyhow::Result. We should probably have a discussion/argument on discord as to which is better. I generally prefer the use in simple things like this.

zephyr-build/src/lib.rs
```diff
@@ -12,30 +12,47 @@
...
+
 /// Export boolean Kconfig entries.  This must happen in any crate that wishes to access the
 /// configuration settings.
-pub fn export_bool_kconfig() {
+pub fn export_kconfig_bool_options() {
     let dotconfig = env::var("DOTCONFIG").expect("DOTCONFIG must be set by wrapper");
 
     // Ensure the build script is rerun when the dotconfig changes.
     println!("cargo:rerun-if-env-changed=DOTCONFIG");
     println!("cargo-rerun-if-changed={}", dotconfig);
```
d3zd3z, line 50: I have seen some behavior that also suggests that this construct isn't working. Not really related to this change, but there have been times I have changed a kconfig but found that the values seen in rust didn't change.

zephyr-sys/build.rs
```diff
@@ -11,14 +11,12 @@
 // This builds a program that is run on the compilation host before the code is compiled.  It can
 // output configuration settings that affect the compilation.
 
-use anyhow::Result;
```
d3zd3z, line 14: As above, I tend to prefer the use of `anyhow::Result`. Are there general guidelines on this?

zephyr/build.rs
```diff
@@ -12,6 +12,6 @@
 // output configuration settings that affect the compilation.
 
 fn main() {
-    zephyr_build::export_bool_kconfig();
+    zephyr_build::export_kconfig_bool_options();
```
d3zd3z, line 15: This is a breaking API change to zephyr_build. This really makes me think we shouldn't be trying to match the crate version numbers with the zephyr numbers, as Zephyr doesn't do semantic versioning (at all). If our crates were all a "0.*" version, the API changes wouldn't matter.

zephyr-build/src/lib.rs
```diff
@@ -12,30 +12,47 @@
...
+/// Extract boolean Kconfig entries.  This must happen in any crate that wishes to access the
+/// configuration settings.
+pub fn extract_kconfig_bool_options(path: &str) -> anyhow::Result<HashSet<String>> {
+    let config_y = Regex::new(r"^(CONFIG_.*)=y$").expect("hardcoded regex is always valid");
+
+    let input = File::open(path)?;
+    let flags: HashSet<String> =
+        BufReader::new(input)
+            .lines()
+            .fold(HashSet::new(), |mut set, line| {
```
d3zd3z, line 31: Using fold with a mutable set feels, well, weird to me. Not strongly enough to ask you to change it. But, I would have just used a loop to make it clear the thing was mutating a single set.

Now, on the other hand:
```rust
.lines()
.filter(|line| config_y.is_match(line))
.collect()
```
is even more concise, well with the error handling fix. In general, I tend to find as things like this get more complicated, the comprehension versions can get a bit unwieldy.

d3zd3z (commented):
Looks good. Thanks for this contribution.

zephyr-sys/build.rs
jpeeters: I use `anyhow::Result` with the fully qualified path in order to prevent confusion from using `::core::result::Result`. This a practice I see several time in different project but we can change to what you think is better.

zephyr/build.rs
jpeeters: Yes this is definitely a breaking change. I suppose that the the project (i.e. `zephyr-lang-rust`) is young enough to allow with breakages at the moment.

I agree that if Zephyr is not using semver, we should not try to match the Zephyr version but have a compatibility table.

zephyr-build/src/lib.rs
jpeeters: I understand. I think your proposal is more readable. I'll make the change.

zephyr/build.rs
d3zd3z: I'll make a PR that reverses the Zephyr version number matching, going back to the 0.x versions, until we get some stability.

d3zd3z:
@inthehack, sorry this a bit neglected. Do you mind trying to resolve the merge conflicts, and pushing a rebased version? It might also need some extension as the bindings have grown a bit.

PR #42, opened 2024-12-12
zephyr: sys: Create Timer

d3zd3z:
Add a set of wrappers for Zephyr's timer.  Timers start as `StoppedTimer`, which can be either dynamic or static (through object). This can then be started either with `start_simple` or `start_callback` which will return a `SimpleTimer` or a `CallbackTimer`.  This can then be turned back into stopped timers with a `stop` method.

### DIFF
--- a/zephyr/src/timer.rs
+++ b/zephyr/src/timer.rs
@@ -1,0 +1,385 @@
+// Copyright (c) 2024 EOVE SAS
+// Copyright (c) 2024 Linaro LTD
+// SPDX-License-Identifier: Apache-2.0
+
+//! Zephyr timers
+//!
+//! This provides a relatively high-level and almost safe interface to Zephyr's underlying
+//! `k_timer`.
+//!
+//! Every timer starts as a [`StoppedTimer`], which has been allocated, but is not tracking any
+//! time.  These can either be created through [`StoppedTimer::new`], or by calling `.init_once(())` on a
+//! StaticStoppedTimer declared within the `kobject_define!` macro, from [`object`].
+//!
+//! The `StoppedTimer` has two methods of interest here:
+//!
+//! - [`start_simple`]: which starts the timer.  This timer has methods for robustly getting counts
+//!   of the number of times it has fired, as well as blocking the current thread for the timer to
+//!   expire.
+//! - [`start_callback`]: which starts the timer, registering a callback handler.  This timer will
+//!   (unsafely) call the callback function, from IRQ context, every time the timer expires.
+//!
+//! Both of these returned timer types [`SimpleTimer`] and [`CallbackTimer`] have a [`stop`] method
+//! that will stop the timer, and give back the original `StoppedTimer`.
+//!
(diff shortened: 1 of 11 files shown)


cfriedt (approved):
LGTM - in the test, I originally mistook 'simples' for 'samples' and thought it was a typo. Naming things is hard 🤷🏼‍♂️

zephyr/src/timer.rs
```diff
@@ -0,0 +1,362 @@
...
+
+    /// The callback used for expiry.
+    expiry: Callback<T>,
+
+    /// Marker to prevent unpinning.
+    _marker: PhantomPinned,
+}
+
+impl<T: Send + Sync> CallbackTimer<T> {
+    fn new(item: Fixed<k_timer>,
```
Ablu, line 297: Maybe take a `StoppedTimer` here to make the safety constraint more explicit? It would also force the consumption of the `StoppedTimer` so it seems like an improvement in robustness?

zephyr/src/timer.rs
```diff
@@ -0,0 +1,362 @@
...
+}
+
+impl Drop for SimpleTimer {
+    fn drop(&mut self) {
+        // Stop the timer, discarding the inner item.
+        let _ = self.raw_stop();
+    }
+}
+
+/// A timer callback.  The function will be called, in IRQ context being passed the given data.
```
Ablu, line 264: > The function will be called, in IRQ context

The comma seems weird here... But my English grammar is pretty bad! So I will leave it to you.

zephyr/src/timer.rs
```diff
@@ -0,0 +1,362 @@
...
+        let item = self.item.take();
+        if let Some(ref item) = item {
+            unsafe {
+                // SAFETY: Stopping the timer removes it from any queues.  There must not be threads
+                // blocked, which is enforced by this only being called from either `stop` or
+                // `drop`.  Once this has been stopped, it is then safe to remove the callback.  As
+                // there will be no more timer operations until the timer is restarted, which will
+                // have a barrier, this is also safe.
+                k_timer_stop(item.get());
+                (*item.get()).expiry_fn = None;
```
Ablu, line 372: nit: `let item = item.get()` before this might be slightly easier to read.

Ablu (approved):
Looks good overall, just some minor comments. Going approve regardless.

I think some of the intermediate commits look like they could be squashed? They just read like minor followups of the main commits.

zephyr/src/timer.rs
d3zd3z: It is permissible either way, but means subtly different things, and I'm not actually sure what I want. I think it probably makes more sense without the comma, though.

d3zd3z:
> I think some of the intermediate commits look like they could be squashed? They just read like minor followups of the main commits.

So, I'm not sure the right answer here. When commits are amended, github basically discards the history and it is really hard to tell what was changed, so I tend to add more commits. For new code like this, it probably makes sense to squash the history down after the reviews are done.

I'll fix the things you mention, which are pretty minor, and then simplify the history..

PR #45, opened 2024-12-20
Work Queue support with Async

d3zd3z:
Zephyr work queue support.  This initial version focuses on using work queues to implement a basic async executor for Zephyr. This is able to schedule work using `spawn`, provides a `spawn_local` function that can spawn from a Future a task to run on the same work queue (allowing non-Send types, such as Rc). The primitives Semaphore, and Queue (and hence channels) have async versions of the blocking methods, and they are fully compatible with the regular Zephyr primitives. There is also a special Mutex implementation that usable entirely from the async context (it is implemented with a Semaphore).


zephyr/src/work.rs
```diff
@@ -0,0 +1,613 @@
...
+//! ## The work queues themselves
+//!
+//! Workqueues themselves are built using [`WorkQueueBuilder`].  This needs a statically defined
+//! stack.  Typical usage will be along the lines of:
+//! ```rust
+//! kobj_define! {
+//!   WORKER_STACK: ThreadStack<2048>;
+//! }
+//! // ...
+//! let main_worker = Box::new(j
```
theotherjimmy, line 136: Extraneous `j`  (do you perhaps use vim ;) 

doc tests might catch this, if that's possible.

zephyr/src/work.rs
```diff
@@ -0,0 +1,613 @@
...
+///
+/// This assumes that Zephyr's works queues have a 1:1 mapping between the work queue and the
+/// thread.
+///
+/// # Safety
+///
+/// The work queue is protected with a sync Mutex (which uses an underlying Zephyr mutex).  It is,
+/// in general, not a good idea to use a mutex in a work queue, as deadlock can happen.  So it is
+/// important to both never .await while holding the lock, as well as to make sure operations within
+/// it are relatively fast.  In this case, `jnsert` and `get` on the SimpleTls are reasonably fast.
```
theotherjimmy, line 329: s/jnsert/insert

zephyr/src/work.rs
```diff
@@ -0,0 +1,613 @@
...
+
+// SAFETY: Zephyr's API maintains thread safety.
+unsafe impl Send for Signal {}
+unsafe impl Sync for Signal {}
+
+impl Signal {
+    /// Create a new `Signal`.
+    ///
+    /// The Signal will be in the non-signaled state.
+    pub fn new() -> Result<Signal, Infallible> {
```
theotherjimmy, line 386: Did I miss why this must be a `Result<Self, Infallible>`? It seems like `Self` would also work here.

zephyr/src/work/futures.rs
```diff
@@ -0,0 +1,605 @@
...
+            let this = this
+                .cast::<u8>()
+                .sub(mem::offset_of!(Self, context))
+                .cast::<Self>();
+            &mut *this
+        }
+    }
+
+    /// Add an event that represents waiting for a semaphore to be available for "take".
+    pub unsafe fn add_semaphore<'a>(&'a mut self, sem: &'a Semaphore) {
```
theotherjimmy, line 288: It's not clear from the documentation what safety invariant the caller must uphold.

zephyr/src/work/futures.rs
```diff
@@ -0,0 +1,605 @@
...
+                ev.get(),
+                ZR_POLL_TYPE_SEM_AVAILABLE,
+                k_poll_modes_K_POLL_MODE_NOTIFY_ONLY as i32,
+                sem.item.get() as *mut c_void,
+            );
+        }
+    }
+
+    /// Add an event that represents waiting for a signal.
+    pub unsafe fn add_signal<'a>(&'a mut self, signal: &'a Signal) {
```
theotherjimmy, line 304: Document caller requirements.

zephyr/src/work/futures.rs
```diff
@@ -0,0 +1,605 @@
...
+                ev.get(),
+                ZR_POLL_TYPE_SIGNAL,
+                k_poll_modes_K_POLL_MODE_NOTIFY_ONLY as i32,
+                signal.item.get() as *mut c_void,
+            );
+        }
+    }
+
+    /// Add an event that represents waiting for a queue to have a message.
+    pub unsafe fn add_queue<'a>(&'a mut self, queue: &'a Queue) {
```
theotherjimmy, line 320: here too.

zephyr/src/work.rs
```diff
@@ -0,0 +1,613 @@
...
+/// let wq = Arc::new(WorkQueueBuilder::new().start(...));
+/// // If the Arc is used after this:
+/// let _ = Arc::into_raw(wq.clone());
+/// // If the Arc is no longer needed:
+/// let _ = Arc::into_raw(wq);
+///
+/// // Leak a work queue in a Box.
+/// let wq = Box::new(WorkQueueBuilder::new().start(...));
+/// let _ = Box::leak(wq);
+///
```
theotherjimmy, line 303: close the code block.

theotherjimmy (changes requested):
nits and requests for unsafe docs, mostly.

zephyr/src/kio.rs
```diff
@@ -0,0 +1,96 @@
...
+        .set_name(name)
+        .start(future)
+}
+
+/// Run an async future on the current worker thread.
+///
+/// Arrange to have the given future run on the current worker thread.  The resulting `JoinHandle`
+/// has `join` and `join_async` methods that can be used to wait for the given thread.
+///
+/// The main use for this is to allow work threads to use `Rc` and `Rc<RefCell<T>>` within async
```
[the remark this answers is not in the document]
d3zd3z: I believe the Send of the output is independent of the Send of the Future itself. At least if I did it right, the async wait on the Answer should properly have Send based on the output.

zephyr/src/work.rs
d3zd3z: Eventually, perhaps. I'm looking into how regular tests might be possible (at least with nightly) as we can provide our own test runner. The doc tests are generally just compiled though, so that might not be as difficult, but is still likely to require nightly.

theotherjimmy: I think doc tests are on stable, and should work during `cargo test`.

teburd (approved):
It's a fairly large PR so I can't say I looked *deeply* at this, but the samples look fine to me and the async bits seem ok, I wouldn't label this kio though myself as Zephyr *has* an async I/O framework already in it that closely matches io_uring called rtio as I've probably noted before.

zephyr/src/work.rs
d3zd3z: More about trying to maintain compatibility with other constructors, but realistically, it probably makes more sense to just have it return Self.

zephyr/src/work/futures.rs
d3zd3z: This has been rearranged, and these functions are now safe. The main safety concern with all of this is that these async methods can only be used with this executor. When I work to get another executor to be available, I'll try to address that.

zephyr/src/work/futures.rs
d3zd3z: Ditto, now safe.

zephyr/src/work/futures.rs
d3zd3z: Also now safe.

samples/work-philosophers/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+crate-type = ["staticlib"]
+
+[dependencies]
+zephyr = "0.1.0"
+
+# Dependencies that are used by build.rs.
+[build-dependencies]
+zephyr-build = "0.1.0"
+
+[profile.release]
```
jpeeters, line 22: Same as above about `codegen-units`.

zephyr/Cargo.toml
```diff
@@ -20,6 +20,7 @@ cfg-if = "1.0"
 
 # The log create is used if the user desires logging, and calls `set_logger()`.
 log = "0.4.22"
+arrayvec = { version = "0.7.6", default-features = false }
```
jpeeters, line 23: Is `arrayvec` better than containers in `heapless`? Or is it for a different use case?

zephyr/src/kio.rs
```diff
@@ -0,0 +1,183 @@
...
+
+pub mod sync;
+
+pub use crate::work::futures::sleep;
+
+/// Run an async future on the given worker thread.
+///
+/// Arrange to have the given future run on the given worker thread.  The resulting `JoinHandle` has
+/// `join` and `join_async` methods that can be used to wait for the given thread.
+pub fn spawn<F>(future: F, worker: &WorkQueue, name: &'static CStr) -> JoinHandle<F>
```
jpeeters, line 29: So as to limit coupling with C types, I think it is more relevant to use `&'static str` here. That being said, type coercion may occur in private functions.

zephyr/src/kio.rs
```diff
@@ -0,0 +1,183 @@
...
+///
+/// To be able to use tasks running on different workers, sharing must be done with types such as
+/// `Arc`, and `Arc<Mutex<T>>`, or `&'static Mutex<T>`.
+///
+/// It is important, when using RefCell, that a borrow from the cell not be carried across an await
+/// boundary, or RefCell's runtime multi-borrow check can cause a panic.
+///
+/// # Panics
+/// If this is called other than from a worker task running on a work thread, it will panic.
+pub fn spawn_local<F>(future: F, name: &'static CStr) -> JoinHandle<F>
```
jpeeters, line 57: Same here.

zephyr/src/kio/sync.rs
```diff
@@ -0,0 +1,125 @@
...
+use crate::{
+    sync::{LockResult, TryLockError, TryLockResult},
+    sys::sync::Semaphore,
+    time::{Forever, NoWait},
+};
+
+/// A mutual exclusion primitive useful for protecting shared data.  Async version.
+///
+/// This mutex will block a task waiting for the lock to become available.
+pub struct Mutex<T: ?Sized> {
```
jpeeters, line 34: Sorry for my misunderstanding. What is the semantic difference with `zephyr::sync::Mutex`?

zephyr/src/lib.rs
```diff
@@ -13,13 +13,19 @@
 pub mod align;
 pub mod device;
 pub mod error;
+#[cfg(CONFIG_RUST_ALLOC)]
+pub mod kio;
```
jpeeters, line 17: A general question: would it be possible to have a `no_alloc` version with `heapless` or do you think that wouldn't be relevant?

zephyr/src/simpletls.rs
```diff
@@ -0,0 +1,124 @@
...
+///
+/// This will maintain a mapping between Zephyr threads and a value of type T.  Entries will have to
+/// be added manually, generally when each thread is started.
+///
+/// Note that T must implement Copy, as it is not safe to retain references to the inner data
+/// outside of this api.
+///
+/// T must also implement Send, since although 'get' always retrieves the current thread's data,
+/// `insert` will typically need to move `T` across threads.
+pub struct SimpleTls<T: Copy + Send> {
```
jpeeters, line 26: I understand the need for not too long name. Although, I think this could be helpful for the developer experience to have something more explicit like `SimpleLocalStorage`. Even more, we can agree that a simple implementation could be named `ThreadLocalStorage` and more advanced ones with specific names. In the end, if some developer prefer shorter names, they can alias to `Tls` internally in their project.

zephyr/src/simpletls.rs
```diff
@@ -0,0 +1,124 @@
...
+///
+/// The StaticTls type has a constant constructor, and the same insert and get methods as the
+/// underlying SimpleTls, with support for initializing the Mutex as needed.
+// TODO: This should eventually make it into a more general lazy mechanism.
+pub struct StaticTls<T: Copy + Send> {
+    /// The container for the data.
+    ///
+    /// The AtomicPtr is either Null, or contains a raw pointer to the underlying Mutex holding the
+    /// data.
+    data: AtomicPtr<Mutex<SimpleTls<T>>>,
```
jpeeters, line 69: I don't understand why we need an atomic on a `Sync` type like `Mutex` here. If this is because of the null possible value, maybe the type could be `Mutex<MaybeUninit<SimpleTls<T>>>` or something equivalent?

jpeeters (changes requested):
Thank you so much for such addition to Rust bindings. I look forward to looking at the next steps.

I set this review as `request changes`, sorry I don't want to be rude, especially at one day of code freeze. Feel free to bypass this time if you need it.

samples/bench/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+[dependencies]
+zephyr = "0.1.0"
+critical-section = "1.1.2"
+
+# Dependencies that are used by build.rs.
+[build-dependencies]
+zephyr-build = "0.1.0"
+
+[profile.release]
+debug-assertions = true
```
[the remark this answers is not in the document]
d3zd3z: I can give it a try.

samples/bench/src/basesem.rs
```diff
@@ -0,0 +1,117 @@
...
+// SAFETY: These are both Send and Sync. The semaphore itself is safe, and the atomic+critical
+// section protects the state.
+unsafe impl Send for Semaphore {}
+unsafe impl Sync for Semaphore {}
+
+impl Semaphore {
+    /// Construct a new semaphore, with the given initial_count and limit.  There is a bit of
+    /// trickery to pass the initial values through to the initializer, but otherwise this is just a
+    /// basic initialization.
+    pub fn new(initial_count: c_uint, limit: c_uint) -> Semaphore {
```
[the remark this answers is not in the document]
d3zd3z: In this case, this was intended to just be a quick test to see if the "base sem" model will be workable. Given that this has similar performance to using `Fixed`, I will likely make a larger change across the codebase to have the zephyr objects directly, with an init atomic.  I won't be changing the APIs to the existing types.

samples/bench/src/basesem.rs
```diff
@@ -0,0 +1,117 @@
...
+            let ptr = this.item.get();
+            (*ptr).count = initial_count;
+            (*ptr).limit = limit;
+        }
+
+        this
+    }
+
+    /// Get the raw pointer, initializing the `k_sem` if needed.
+    fn get(&self) -> *mut k_sem {
```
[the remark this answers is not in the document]
d3zd3z: The point of this little exercise (basesem, which will go away at some point) is to see about the feasibility of eliminating the `Fixed` type entirely. In this case, the items are always just used directly, and the atomic protects against things being moved (which only happens when there is a single reference anyway).

samples/bench/src/basesem.rs
```diff
@@ -0,0 +1,117 @@
...
+            }
+
+            self.state
+                .store(self as *const Self as usize, Ordering::Release);
+            self.item.get()
+        })
+    }
+
+    /// Synchronous take.
+    pub fn take(&self, timeout: impl Into<Timeout>) -> Result<()> {
```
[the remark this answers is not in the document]
d3zd3z: At least for semaphores, I was trying to keep the same names as Zephyr. Embassy, for example, uses acquire and release. The very traditional names would be "try" and "increment", or even better, the Dutch "probeer" and "verhoog".

Names are always a little weird with semaphores because they tend to come from a specific way that semaphores are being used, whereas semaphores are actually pretty generic.

zephyr/src/kio.rs
d3zd3z: Honestly, I'm not quite sure what the best thing to do is. This `name` _must_ be null terminated, and must be static, so other than by leaking pointers or some kind of allocation, it has to be a CStr.

Realistically, I wonder if we should have the name stuff in here at all. I was hoping to be able to get the name passed through to SystemView, but it never ended up working, and otherwise, there isn't much purpose to it.

zephyr/src/kio/sync.rs
d3zd3z: Zephyr's Mutex type is only blocking, and using from async would block the entire work queue. This is kind of an interim solution to provide mutual exclusion (without condvar) that works from the workqueue-based async.

zephyr/src/lib.rs
d3zd3z: I think it'd be a good idea, and a good thing, too, and a lot more reasonable to do after I get rid of `Fixed`. But, I think that is better to do as some changes to come after this.

zephyr/src/simpletls.rs
d3zd3z: Perhaps `pub(crate)`? It really isn't intended for general consumption.

zephyr/src/simpletls.rs
d3zd3z: The problem is that currently, `Mutex` cannot be declared statically (the constructor is not const), as it requires an allocation. My proposal suggested in basesem will fix that.

d3zd3z:
> Thank you so much for such addition to Rust bindings. I look forward to hearing from you on the next steps.
> 
> I set this review as `request changes`, sorry I don't want to be rude, especially at one day of code freeze. Feel free to bypass this time if you need it.

Given the closeness to the feature freeze, I'm going to take the above as a kind of ok, to proceed with this, and make the changes in subsequent PRs. I think it will help to have this code in mainline for this release.

PR #49, opened 2025-01-10
zephyr: time: Add "now" for Instant

aolowin:
Uses the full tick resolution rather than a millisecond conversion.


theotherjimmy (approved):

d3zd3z:
Thanks for your contribution. I did already add this in the "work q" PR, but I will add an acknowledgement to your contribution there. I _think_ you are right in only really providing this function when time is 64-bit, as `Instant` just doesn't make a lot of sense with an overflowing timer.

d3zd3z:
I believe this is now redundant.

PR #54, opened 2025-02-04
CI: Change doc generation to use a more generic target

d3zd3z:
Rather than targeting a specific board, use a Qemu target, which should make the docs fairly general. The generated devicetree will be minimal, but users wishing to view these docs should be generating them for a specific build anyway.


teburd:
typically with rustdoc you'd enable all feature flags, is that sort of thing not possible?

d3zd3z:
> typically with rustdoc you'd enable all feature flags, is that sort of thing not possible?

Well, that's up to the prj.conf in the docgen project. We don't really use features with Zephyr, as they come from Kconfigs.

ithinuel (approved):

cfriedt (approved):
Lgtm

PR #57, opened 2025-02-20
Convert Zephyr primitives to a better initialization and use framework

d3zd3z:
The existing object wrappers for Zephyr objects centers around a "Fixed" type, which is a enum, containing either a static reference, or a `Pin<Box<T>>`. This means that anything not declared with the `kernel_obj` macros will require allocation.

We can get rid of the need for allocation and also improve performance (very slightly). Instead of this enum which then hides another reference, a ZephyrObject is just a regular struct, with an atomic and an UnsafeCell holding the real Zephyr object. The atomic is used to allow the initialization to happen on first use instead of at constructor time. It also detects if one of these objects has been moved (resulting in a panic), because requiring the use of Pin on these types would make this mechanism even less hygienic than the `kernel_obj` macros.

The new objects have const constructors, which allows them to be plainly declared as statics, or used in regular code, without requiring an alloc to happen under the hood. The only constraint is that the objects cannot be moved after their first use. For embedded, this is generally not difficult, as the objects will usually be static, put in something like a StaticCell, or allocated in an Rc or Arc.

There is, unfortunately, a bit of a minor API change. Semaphore's currently return a Result, and the Result type is not particularly usable in const. `Semaphore::new` now directly returns the Semaphore, and if the constructor values are nonsensical (either a zero limit, or an initial value larger than the limit), it will panic instead of returning an error.


cfriedt (approved):
LGTM - I think it's a bit simpler with fewer clones and unwraps.

Users can also adjust the git ref that the Zephyr 4.1.0 release points to, if they want to try this out with the latest Zephyr release after it's merged.

teburd (approved):

teburd:
I wouldn't be super scared of Pin, but really most IPCs in Zephyr really do want to be static I'd think. Otherwise there's going to be lots of issues trying to track lifetimes of what lives longer than what else. Particularly because Zephyr loves callbacks and there's not a great way I don't think to inform Rust something must live longer than the callback. I could be wrong though!

PR #63, opened 2025-02-28
Initial support for using Embassy with Rust on Zephyr

d3zd3z:
These changes add the ability to use the executor from [Embassy](https://embassy.dev/) on Zephyr. For the most part, Embassy's executor and associated utilities (embassy-time, embassy-sync, embassy-futures) are fairly platform agnostic. This support adds the two things needed to make this all work on Zephyr:

- time-driver. Enabling this feature will use a Zephyr timer to implement the `embassy-timer-driver`, so that all of the time related utilities from `embassy-time` work. There is a bit of challenge here as the embassy utilities expect the tick rate to be set via a feature, and in Zephyr this comes through Kconfig. But, once set correctly, the result should be seamless.
- executor-zephyr. Implements an `Executor` that works just like `embassy-executor`'s built in Executor type, except that it is intended to be run on a single Zephyr thread. More than one Executor can be run, each on its own thread, to provide a useful compromise between the lower-overhead scheduling of an executor, and priority-based scheduling that comes from using multiple threads.

Some simple benchmarking shows the following. All of these are measuring the round trip time between two async tasks sending a single message over a channel, and a result back.
- **12us**: Native Embassy switch time. This is with just the `executor-thread` executor that expects to be running on the single lowest priority thread.
- **15us**: The `executor-zephyr` executor, with all of the async tasks scheduled on a single executor. This includes the above switch time, along with a little overhead needed to support Zephyr threads.
- **26us**: The `executor-zephyr` executor, with one task of the pair running on a separate thread. This includes the zephyr context switch time in with the executor switch time.

All of these scenarios are useful in different situations.

### DIFF
--- a/samples/embassy/CMakeLists.txt
+++ b/samples/embassy/CMakeLists.txt
@@ -1,0 +1,8 @@
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.20.0)
+
+find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
+
+project(embassy_on_zephyr)
+rust_cargo_application()
--- a/samples/embassy/Cargo.toml
+++ b/samples/embassy/Cargo.toml
@@ -1,0 +1,51 @@
+# Copyright (c) 2024 Linaro LTD
+# SPDX-License-Identifier: Apache-2.0
+
+[package]
+# This must be rustapp for now.
+name = "rustapp"
+version = "0.1.0"
+edition = "2021"
+description = "A sample hello world application in Rust"
+license = "Apache-2.0 or MIT"
+
+[lib]
+crate-type = ["staticlib"]
+
+[dependencies]
+zephyr = { version = "0.1.0", features = ["time-driver"] }
+log = "0.4.22"
+static_cell = "2.1"
+
+[dependencies.embassy-executor]
+version = "0.7.0"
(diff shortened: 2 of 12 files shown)


teburd (approved):
Running embassy on Zephyr wasn't on my 2025 bingo sheet, but I have to say this is really quite cool.

I wonder how this might work if you want to run the full embassy network stack in place of Zephyrs for example, where security might be a real honest to goodness concern and Rust would likely help. I kind of love the possibilities this opens up.

d3zd3z:
@teburd a couple of thoughts:

As far as using parts of Embassy instead of Rust. The main challenge here is hooking the device interrupts in so that the call into the hal. Since for most things in the hal, embassy requires this to be done by hand, this shouldn't be all that difficult. But, the mechanisms are very different, and Zephyr is going to need the irqs to still go through it's irq handling, as this will be needed in order for an irq to be able to cause the Zephyr scheduler to do something.

Otherwise, I would think it would then mostly just be an issue of disabling that functionality on the Zephyr side. Running a mix of these shouldn't be all that difficult.

As far as user mode goes, getting all of the permissions working would probably be a bit hard, especially since the hal is going to want access to hardware blocks. If there are enough MPU regions, this is probably doable, although it is a little unclear how user mode will help rust code, as it is largely redundant. It almost seems like more useful would be to have the userspace used to protect the rest of the system from some block of C code.

cfriedt (approved):

PR #65, opened 2025-03-05
Deprecate old async

d3zd3z:
Begin the process of deprecating the old "async" implementation. This implementation has numerous problems:
- It is fundamentally incompatible with how async executors are expected to work. Trying to combine with other libraries that expect async will either result in a `panic`, or even undefined behavior.
- They are only usable with a limited set of scheduling primitives, notably Zephyr's `k_sem`, and `k_queue`. This is largely a limitation of what is supported by triggered in Zephyr.
- They're not useful for much else. In order to work with something like a driver callback, that interface would have to use a Semaphore or Channel, since the Waker from the async call itself cannot be used.
- They aren't very efficient. Using triggered work ends up using the combined overhead of threads, and work queues. Using the utilities provided by Embassy allows same-thread scheduling to be very low overhead, while bringing in the thread scheduling overhead only when wakes will cross thread boundaries.

The proposed alternative is to use the `Executor` provided by the feature `executor-zephyr`. This is a very lightweight executor, making use of some utilities provided by Embassy to allow lightweight scheduling on a single thread, and reasonably efficient scheduling across Zephyr threads. Async device operations can use the `Waker` mechanism as designed to be able to provide nice async Rust interfaces to drivers.

### DIFF
--- a/docgen/Cargo.toml
+++ b/docgen/Cargo.toml
@@ -13,4 +13,4 @@ license = "Apache-2.0 or MIT"
 crate-type = ["staticlib"]
 
 [dependencies]
-zephyr = { version = "0.1.0", features = ["executor-zephyr", "async-drivers"] }
+zephyr = { version = "0.2.0", features = ["executor-zephyr", "async-drivers"] }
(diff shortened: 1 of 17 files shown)


cfriedt (approved):

teburd (approved):
I'd simply drop it unless something depends on this

d3zd3z:
Instead of deprecating these, I'll go ahead and just drop the support.

PR #67, opened 2025-03-07
task + async misc

d3zd3z:
Create a new proc-macro-based way of declaring threads in Zephyr. Instead of the weird `kobj_define` macro, provide a `zephyr::thread` macro that can decorate an ordinary Rust function to become a builder for a `ReadyThread` that can just be started. Threads can also now be reused after they exit.

This allows something like:
```rust
#[zephyr::thread(stack_size = 2048)]
fn mythread(name: &'static str, index: usize) {
    // Body
}
```

and somewhere in the code, you just call it, which returns a handle that can be used to start the thread.
```rust
let joiner = mythread("thename", 42).start();
...
joiner.join();  // Wait for thread to exit, if desired.
```

There is still some things to clean up, hence the draft, but I wanted to get some exposure.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -45,7 +45,8 @@ jobs:
         working-directory: zephyr-lang-rust
         run: |
           # Note that the above build doesn't set Zephyrbase, so we'll need to do that here.
-          west build -t rustdoc -b qemu_cortex_m3 docgen
+          west build -b qemu_cortex_m3 docgen
+          west build -t rustdoc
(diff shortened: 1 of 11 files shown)


d3zd3z:
I'd like to start getting this ready for review. But, there is a question of how much this should be split apart. There are a few cherry-picked changes that are to be merged, so we'll at least wait for those to go in.

d3zd3z:
This ended up getting merged as part of another PR, so closing this PR.

Issue #66, opened 2025-03-06
windows rust build problem


coskunergan:
I using windows 10 and can compile standard zephyr demos. but, cannot compile the sample at "C:\zephyr\zephyrproject\modules\lang\rust\samples\hello_world"  my build command is "west build -b qemu_cortex_m3"

I get the following message in all my attempts.

` thread 'main' panicked at C:\Users\CoskunERGAN\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\bindgen-0.70.1\lib.rs:622:27:
  Unable to find libclang: "the libclang shared library at C:\\msys64\\mingw64\\bin\\libclang.dll could not be opened: LoadLibraryExW failed"
  note: run with RUST_BACKTRACE=1 environment variable to display a backtrace`

I installed cargo in msys64 and installed rustup on it. Also, clang and llvm are installed under the msys64 folder.

Why can't I compile any Rust examples even though the environment variables are written as well?

![Image](https://github.com/user-attachments/assets/763680b5-3a63-4414-a400-e7639e377292)

![Image](https://github.com/user-attachments/assets/f9263df8-7a0c-4b7e-9311-775804bce660)

![Image](https://github.com/user-attachments/assets/1e6ccce8-de9a-4f3f-bfc0-ce239e9a923c)

coskunergan:
similar reasons;

https://github.com/zephyrproject-rtos/zephyr-lang-rust/issues/68

Issue #68, opened 2025-03-10
Rust Hello World build parsing zephyr.dts error


coskunergan:
Hi, I have Clang and llvm and Rust installed. But I can't compile the RUST hello_world demo. 

getting the following error message.

`  thread 'main' panicked at C:\zephyr\zephyrproject\modules\lang\rust\zephyr-build\src\devicetree\parse.rs:23:46:
  Parsing zephyr.dts: Error { variant: ParsingError { positives: [node], negatives: [] }, location: Pos(9), line_col: Pos((1, 10)), path: None, line: "/dts-v1/;ÔÉıÔÉè", continued_line: None, parse_attempts: None }
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace`

/dts-v1/;ÔÉıÔÉè" -> When I checked the zephyr.dts file there is no file corruption mentioned there.

My detailed compilation outputs are attached. "build.log.txt" end "verbose_build_log.txt"

Is there anyone who can help me?

[build_log.txt](https://github.com/user-attachments/files/19158516/build_log.txt)

[verbose_build_log.txt](https://github.com/user-attachments/files/19158519/verbose_build_log.txt)

coskunergan:
I solved the problem.

Adding the "\r" character to the content of this file solves the compilation problem for Windows.

"../modules/lang/rust/zephyr-build/src/devicetree/dts.pest" line: 77

`WHITESPACE` = _{ " " | "\r" | "\n" | "\t" }` 

is now compiling to RUST.

kartben:
@coskunergan will you be able to send a pull request to get this fixed for everyone? :)

coskunergan:
@kartben I did it. ;)

https://github.com/zephyrproject-rtos/zephyr-lang-rust/pull/69

PR #69, opened 2025-03-13
Added "\r" to WHITESPACE char for windows build.

coskunergan:
see: https://github.com/zephyrproject-rtos/zephyr-lang-rust/issues/68

### DIFF
--- a/zephyr-build/src/devicetree/dts.pest
+++ b/zephyr-build/src/devicetree/dts.pest
@@ -74,4 +74,4 @@ nodename = @{
     (ASCII_ALPHANUMERIC | "_" | "," | "." | "?" | "-" | "@" | "#")+
 }
 
-WHITESPACE = _{ " " | "\n" | "\t" }
+WHITESPACE = _{ " " | "\r" | "\n" | "\t" }


zephyr-build/src/devicetree/dts.pest
```diff
@@ -74,4 +74,4 @@ nodename = @{
     (ASCII_ALPHANUMERIC | "_" | "," | "." | "?" | "-" | "@" | "#")+
 }
 
-WHITESPACE = _{ " " | "\n" | "\t" }
+WHITESPACE = _{ " " | "\r" | "\n" | "\t" }
```
d3zd3z, line 77: This definitely will work, although I think technically it should be "\r\n" and "\n", but this solution will also allow this to run on ancient Macs as well.

d3zd3z (changes requested):
A few minor issues, just with the commit text:

1. The commit needs to be signed off by. This is inherited from the Zephyr project, although we should probably come up with guidelines for this project.
2. There should be a bit more description in the commit text. Linking to the GH issue is fine (although just `Fixes: #66` is better, as it will auto-fix when this merges).
3. There should be a short tag in front of the title. Something like "zephyr-build: " is fine.
4. Lastly, even more minor, the commit text should be in the present tense. Using the past tense probably feels a little more natural when writing, but the commits persist, and are generally read in the context of history.

What I'd suggest is something like:

```
zephyr-build: fix DTS parser for Windows

The zephyr-build DTS parser fails on Windows, due to the "\r\n" whitespace used on that platform. Accept the "\r" in addition to the "\n" to allow it to parse successfully.

Fixes: #66

Signed-off-by: Firstname Lastname <coskungeran@gmail>
```

I could update this for you if you'd prefer, but I do need to know what fullname to use.

d3zd3z:
CI issue with docs is unrelated.

d3zd3z (approved):

PR #77, opened 2025-04-01
Add rust-toolchain file

d3zd3z:
This specifies the exact tools that need to be present in order to build these projects.  This should affect all of the samples.

Currently, we require Rust 1.85 for a few reasons.  There have been some subtle changes to warnings that cause build failures with older version. Specifically, the 2024 edition requires certain attributes to explicitly be marked as unsafe, but the construct to mark them as unsafe is only present in the 2024 compiler.  There are also some subtle changes to warnings that affect our builds.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,12 @@
+[toolchain]
+channel = "1.85"
+components = [ "rust-src", "rustfmt", "llvm-tools" ]
+targets = [
+    "thumbv7em-none-eabi",
+    "thumbv7m-none-eabi",
+    "thumbv6m-none-eabi",
+    "thumbv7em-none-eabihf",
+    "thumbv8m.main-none-eabihf",
+    "riscv32i-unknown-none-elf",
+    "riscv64imac-unknown-none-elf",
+]


rust-toolchain.toml
```diff
@@ -0,0 +1,12 @@
+[toolchain]
+channel = "1.85"
+components = [ "rust-src", "rustfmt", "llvm-tools" ]
+targets = [
```
teburd, line 4: It would be great to see some additional architectures being tested for here...

x86, xtensa (esp32), sparc, and mips are all supported by upstream rust and by zephyr
https://doc.rust-lang.org/rustc/platform-support/xtensa.html

Not a blocker, but something I'd love to see

teburd (approved):

d3zd3z:
So, we need to hold off on this, and perhaps put some thought into this. The CI image has the rust toolchain installed as root, and if cargo thinks it needs to download tools, it will just fail with permission errors.

Perhaps a better solution will be to just specify a minimum rust version in the Cargo files. It will still fail when CI's tools are out of date, but in a more obvious way. This will also keep people happy that don't want to see zephyr's CI doing things like downloading toolchains.

d3zd3z:
Maybe this will return later. Currently, if the version requested isn't the same as the one in the Zephyr CI, cargo just fails trying to upgrade the tools.

PR #78, opened 2025-04-02
cmake: add USES_TERMINAL to show cargo output

dario23:
Otherwise the CMake output is just stuck at the rust build step for
however long it takes to build.

Signed-Off-By: Johannes Schilling <github@deaktualisierung.org>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,6 +157,7 @@ ${config_paths}
   add_custom_command(
     OUTPUT ${DUMMY_FILE}
     BYPRODUCTS ${RUST_LIBRARY} ${WRAPPER_FILE}
+    USES_TERMINAL
     COMMAND
       ${CMAKE_COMMAND} -E
       env BUILD_DIR=${CMAKE_CURRENT_BINARY_DIR}
@@ -199,6 +200,7 @@ ${config_paths}
   # it is important to keep this in sync with the above.
   add_custom_command(
     OUTPUT generate_rust_docs
+    USES_TERMINAL
     COMMAND
       ${CMAKE_COMMAND} -E
       env BUILD_DIR=${CMAKE_CURRENT_BINARY_DIR}


dario23:
@d3zd3z this is what should fix the issue we briefly touched on in today's Zephyr Rust status talk; marking it as draft because i can't test with my current environment.

CMakeLists.txt
```diff
@@ -157,6 +157,7 @@ ${config_paths}
   add_custom_command(
     OUTPUT ${DUMMY_FILE}
     BYPRODUCTS ${RUST_LIBRARY} ${WRAPPER_FILE}
+    USES_TERMINAL
```
d3zd3z, line 160: This is good, but it should be added to both `add_custom_command` blocks, so that document generation will also be verbose.

d3zd3z (changes requested):
If you get a chance, can you add to the other add custom command block. Also, the commit itself should have a `Signed-off-by` block at the end. See the other commits for more information. I think Zephyr guidelines also limit the length of lines in the commit text to 75 or so.

d3zd3z:
Let me know, if you prefer, and I can just pull this change into some upcoming changes I have. I will need to have your full name for the email and the signed-off-by.

Or feel free to update it yourself.

dario23:
updated; also changed the commit description/first comment here, not sure what ends up in the commits in the end.

d3zd3z:
I should try to get checkpatch running. @dario23 this is the current output. These should be easy to fix. You'll have to do:
```
$ git config user.name 'Johannes Schilling'
$ git commit --amend --reset-author
```
to get it to update the author. The rest are the styling and formatting.

You can use `git commit ... -s` to add the signed-off-by line automatically (it works with amend).

```
WARNING: Do not use whitespace before Signed-off-by:
#10: 
    Signed-Off-By: Johannes Schilling <github@deaktualisierung.org>

WARNING: 'Signed-off-by:' is the preferred signature form
#10: 
    Signed-Off-By: Johannes Schilling <github@deaktualisierung.org>

WARNING: Missing Signed-off-by: line by nominal patch author ''

total: 0 errors, 3 warnings, 14 lines checked

NOTE: For some of the reported defects, checkpatch may be able to
      mechanically convert to the typical style using --fix or --fix-inplace.

Your patch has style problems, please review.

NOTE: If any of the errors are false positives, please report
      them to the maintainers.
```

It is complaining that the author doesn't match the signed off by (which can be fixed with the reset author), and about the capital O and B in the Sign-off-by.

d3zd3z (approved):

cfriedt (approved):
Looks OK to me - from what I gather, this should show incremental build progress as the rust compiler builds sources, right?

teburd (approved):

PR #79, opened 2025-04-03
zephyr:embassy: Use a semaphore for the executor

aolowin:
Fixes a potential race condition in the suspend/resume sequence.

### DIFF
--- a/zephyr/src/embassy.rs
+++ b/zephyr/src/embassy.rs
@@ -48,10 +48,9 @@
 //!
 //! The following features in the `zephyr` crate configure what is supported:
 //!
-//! - **`executor-zephyr`**: This implements an executor that uses Zephyr's thread primitives
-//!   (`k_thread_suspend` and `k_thread_resume`) to suspend the executor thread when there is no work
-//!   to perform.  This feature is incompatible with either `embassy-thread` or `embassy-interrupt`
-//!   in the `embassy-executor` crate.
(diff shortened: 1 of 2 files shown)


d3zd3z:
@aolowin I've approve the workflow to run, so you should see the rustfmt fixes that will be needed.

Also, you will need a Signed-off-by line in the commit, which you should be able to add with `git commit --amend -s`.

The other minor nit is that there should be a space after the colon in the commit title, e.g. `zephyr: embassy: ...`.

aolowin:
I'm not sure why the build is failing on Semaphore::new(0, 1).unwrap()

The unwrap() is needed on my build to compile and it's used in a few of the samples.

d3zd3z (approved):
My testing shows this to work well. I like the solution, and this will make it easier to add support for async wait on other semaphores.

cfriedt (approved):
LGTM

d3zd3z:
> I'm not sure why the build is failing on Semaphore::new(0, 1).unwrap()
> 
> The unwrap() is needed on my build to compile and it's used in a few of the samples.

This was unfortunately needed to allow semaphores to be declared as statics, which needs the construct to be const.  Result is technically returnable from const, but it is almost impossible to use (at least with stable rust). So, I changed the behavior to panic if the initializer is nonsensical, and just return the Semaphore.

PR #80, opened 2025-04-04
Add support for board native_sim/native/64

d3zd3z:
This allows the native_sim/native/64 target to build on both x86_64 and aarch64 Linux.

Currently 32-bit native_sim is not supported, as upstream rust does not build std/core/alloc for the unknown-none variants of these targets.

This adds the "zephyr,sim-flash" device so that the generated DT will compile.

Based on a contribution by:

    Author: Evan Perry Grove <evan@4grove.com>
    Date:   Wed Jan 29 13:00:29 2025 -0600


d3zd3z:
@ryukoposting I've migrated your native sim patch to this module, where Rust on Zephyr is being developed.

cfriedt (approved):

teburd (approved):

PR #81, opened 2025-04-09
zephyr: Replace critical-section implementation

d3zd3z:
There is a fairly fundamental incompatibility between Zephyr spin locks and the Critical Section specification.  Zephyr spin locks do not allow nesting from within a single spin lock.  The critical section API only has an `acquire` and `release` entry, and provides no way (such as a stack frame) to have a unique context for different invocation places.

Unfortunately, this means we cannot use spin locks for critical sections.

Instead, this change implements critical sections using irq locking. The implementation of these macros on Zephyr does try to make them SMP safe, with a simple atomic lock, but there is still something preventing the riscv SMP from working.

Also, these entries cannot be called from user mode.  There are various other reasons we don't support usermode, so at this time, just have a compile time assertion that usermode is not enabled in the build.  If it is needed, we will have to come up with another way to implement this.

### DIFF
--- a/zephyr-sys/build.rs
+++ b/zephyr-sys/build.rs
@@ -77,6 +77,7 @@ fn main() -> Result<()> {
         .allowlist_function("k_.*")
         .allowlist_function("gpio_.*")
         .allowlist_function("flash_.*")
+        .allowlist_function("zr_.*")
         .allowlist_item("GPIO_.*")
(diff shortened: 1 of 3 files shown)


cfriedt (approved):

teburd (approved):
Lgtm

PR #87, opened 2025-04-11
Backport work queue builder

bjarki-andreasen:
Backport work queue builder, see https://github.com/zephyrproject-rtos/zephyr-lang-rust/pull/86

### DIFF
--- a/ci-manifest.yml
+++ b/ci-manifest.yml
@@ -12,7 +12,7 @@ manifest:
   projects:
     - name: zephyr
       remote: zephyrproject-rtos
-      revision: main
+      revision: pull/88345/head
       import:
         # By using name-allowlist we can clone only the modules that are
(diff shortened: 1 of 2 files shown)


d3zd3z:
Want to make sure this gets a DNM, as this definitely cannot be merged until the referenced PR is merged into Zephyr. Even then, we have to be careful because the Rust code is frequently used with earlier versions of Zephyr.

bjarki-andreasen:
> Want to make sure this gets a DNM, as this definitely cannot be merged until the referenced PR is merged into Zephyr. Even then, we have to be careful because the Rust code is frequently used with earlier versions of Zephyr.

This is not possible though, given zephyr is running CI building this module, so they have to be merged together. If we remove rust from upstream CI, then this can wait, but for now, its blocking zephyr...

PR #90, opened 2025-04-11
zephyr-build: add aux_display support and fixed the gpio interrupt for stm32 series.

coskunergan:
The zephyr-build parser was missing aux_display support, which was included and used in a demo;

https://github.com/coskunergan/zephyr_rust_demo

### DIFF
--- a/zephyr/src/sys.rs
+++ b/zephyr/src/sys.rs
@@ -39,42 +39,42 @@ pub fn uptime_get() -> i64 {
     unsafe { crate::raw::k_uptime_get() }
 }
 
+// The below implementation, based on interrupt locking has only been tested on single CPU.  The
+// implementation suggests it should work on SMP, and can be tested.  The docs for irq_lock()
+// explicitly state that it cannot be used from userspace. Unfortunately, spinlocks have
+// incompatible semantics with critical sections, so to work with userspace we'd need probably a
+// syscall.
+#[cfg(CONFIG_USERSPACE)]
+compile_error!("Critical-section implementation does not work with CONFIG_USERSPACE");
+
 pub mod critical {
     //! Zephyr implementation of critical sections.
     //!
-    //! Critical sections from Rust are handled with a single Zephyr spinlock.  This doesn't allow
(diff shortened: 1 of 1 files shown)


zephyr-sys/build.rs
```diff
@@ -76,6 +76,7 @@ fn main() -> Result<()> {
         .derive_copy(false)
         .allowlist_function("k_.*")
         .allowlist_function("gpio_.*")
+		.allowlist_function("auxdisplay_.*")
```
d3zd3z, line 79: Your editor seems to be inserting tabs instead of spaces. It does seem to suggest that cargo fmt doesn't process the build.rs file.

zephyr/src/device/gpio.rs
```diff
@@ -31,7 +31,7 @@ mod async_io {
     use embassy_sync::waitqueue::AtomicWaker;
     use zephyr_sys::{
         device, gpio_add_callback, gpio_callback, gpio_init_callback, gpio_pin_interrupt_configure,
-        gpio_pin_interrupt_configure_dt, gpio_port_pins_t, GPIO_INT_LEVEL_HIGH, GPIO_INT_LEVEL_LOW,
+        gpio_pin_interrupt_configure_dt, gpio_port_pins_t, GPIO_INT_EDGE_FALLING, GPIO_INT_EDGE_RISING,
```
d3zd3z, line 34: Although the STM32 devices do not have level triggered interrupts, we don't want to remove support for them from targets that do. The async interface is actually really racy with edge triggered interrupts, and will pretty readily end up missing an edge and not firing again.

Honestly, I think supporting wake from stm32 is going to be a bit more work. We can't just have something attach an edge trigger, but would need to leave it attached, to track the state.

zephyr-sys/build.rs
d3zd3z: Aside from that, does auxdisplay have anything to do with this PR?

d3zd3z (changes requested):

PR #92, opened 2025-04-13
feat(cargo): Set rustc-check-cfg for all config options

ball-hayden:
This squashes lint warnings for config options which are not currently enabled.

See https://doc.rust-lang.org/cargo/reference/build-scripts.html\#rustc-check-cfg

### DIFF
--- a/scripts/kconfig-bool-options.py
+++ b/scripts/kconfig-bool-options.py
@@ -1,0 +1,50 @@
+#!/usr/bin/env python3
+
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+import sys
+import argparse
+import json
+
+from itertools import chain
+
+sys.path.insert(0, str(os.path.join(os.environ["ZEPHYR_BASE"], "scripts", "kconfig")))
+from kconfiglib import Kconfig, _T_BOOL, _T_CHOICE
+
+
+def main():
+    args = parse_args()
+
+    kconf = Kconfig(args.kconfig_file)
+    kconf.load_config(args.dotconfig)
+
+    options = {}
+
+    for sym in chain(kconf.unique_defined_syms, kconf.unique_choices):
+        if not sym.name:
+            continue
+
+        if "-" in sym.name:
+            # Rust does not allow hyphens in cfg options.
+            continue
+
+        if sym.type in [_T_BOOL, _T_CHOICE]:
+            options[f"CONFIG_{sym.name}"] = sym.str_value
+
+    with open(args.outfile, "w") as f:
+        json.dump(options, f, indent=2, sort_keys=True)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(allow_abbrev=False)
+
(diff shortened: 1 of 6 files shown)


ball-hayden:
@d3zd3z thanks for your work so far on this.

May I request a review on this, please?

d3zd3z (commented):
Unfortunately, this only squashes _some_ warnings from this lint, specifically settings that are possible in the current configuration, but are disabled. But configurations that don't apply will not be seen by this and will still generate warnings.

Aside from generating a list of all possible configs from all Kconfig files (including modules that aren't present), I don't really see a way to make this work, which is why my examples have just been suppressing the lint.

ball-hayden:
This is true.
It does feel like a step in the right direction, though?

ball-hayden:
I wonder if we could do something with `kconfiglib` similar to how the [Doxygen extension](https://github.com/zephyrproject-rtos/zephyr/blob/main/doc/_extensions/zephyr/kconfig/__init__.py) works, so we could list all possible options.

I'll try to find some time to have a look into this.

ball-hayden:
@d3zd3z how do you feel about this as an approach?

This seems to work for my project - I'd be interested to know how CI fares.

I'm not entirely happy about having an intermediate JSON file, but struggling to think of other options that would produce the complete list we need.

scripts/kconfig-bool-options.py
```diff
@@ -0,0 +1,50 @@
...
+from itertools import chain
+
+sys.path.insert(0, str(os.path.join(os.environ["ZEPHYR_BASE"], "scripts", "kconfig")))
+from kconfiglib import Kconfig, _T_BOOL, _T_CHOICE
+
+
+def main():
+    args = parse_args()
+
+    kconf = Kconfig(args.kconfig_file)
```
d3zd3z, line 19: I'm not familiar with this code from Zephyr. Will this visit all kconfig files, or just the ones that make sense for the current build. Specifically, I think we should try this with a test case of conditional code that references a config that isn't present in the builds we make, but is in the tree. Ideally, we would make sure this works for modules as well.

ball-hayden: > Will this visit all kconfig files, or just the ones that make sense for the current build

My understanding is that this will pick up the applications "root" Kconfig file (i.e. the one in `<appdir>/Kconfig`, where it is defined), and follow any `source`/`rsource` commands from there.

So it's not loading everything that smells like a Kconfig file within the directory tree, but it will load all kconfig files that are `rsource`ed (the same as with Zephyr's normal build).

> a test case of conditional code that references a config that isn't present in the builds we make, but is in the tree

I'm not sure I follow.
If it's loaded as part of the Kconfig tree, it is present? If it isn't part of the Kconfig tree, then it's not a valid option (and should be flagged as not available).

ball-hayden: This and the CMake changes were based around https://github.com/zephyrproject-rtos/zephyr/blob/main/cmake/modules/kconfig.cmake

PR #102, opened 2025-05-29
CI: Use specific version of action-zephyr-setup

d3zd3z:
This should match the version running in the main zephyr repo.


kartben (approved):
For good measure, you'll probably want to pin *all* actions to SHAs

PR #106, opened 2025-06-09
dts: only generate getters for okay nodes

dsseng:
Code generated for disabled nodes produces Rust compiler errors, thus
we must not generate getters for unavailable nodes.

Fixes building for RP2350 (otherwise passes and works).

> A node is considered enabled if its status property
> is either ``"okay"`` or not defined
> For backwards compatibility, the value ``"ok"``
> is treated the same as ``"okay"``

Signed-off-by: Dmitrii Sharshakov <d3dx12.xx@gmail.com>


dsseng:
Examples tested on RP2350: async-philosophers, blinky

d3zd3z (approved):
This looks good, thanks.

PR #108, opened 2025-06-13
Added support for more granular RISCV32/64 ISA extensions

cordawyn:
Rust target for RISCV64 was hard-coded to "riscv64imac-unknown-none-elf" and it failed to link with the rest of Zephyr when some specific ISA extensions were enabled (e.g. `RISCV_ISA_EXT_G`), resulting in errors "can't link soft-float modules with double-float modules". This PR dynamically updates `RUST_TARGET` depending on the enabled extensions. While I was at it, I also did the same for RISCV32 targets.

I did not cover all possible variants, only those currently available in "rustup target list" (which means, only "riscv64gc-unknown-none-elf" and "riscv64imac-unknown-none-elf", as well as a bunch of targets for "riscv32").


d3zd3z (approved):
This does seem to build up the right target, but do we end up in situations where upstream rust hasn't built core/alloc for a given variant? This will get a lot easier when building core/alloc/std just with cargo becomes stable.

d3zd3z:
> So to summarize - I'm pretty new to this project and Zephyr ecosystem, and I'm not sure how we should proceed, given the issues that you've mentioned. For the time being I would be happy if I could somehow convince "west" replace the "official" zephyr-lang-rust with my branch - just for my experiments, at least (until some solution for this PR presents itself?). I've asked about this override in Discord already (in "west" channel), but got no feedback.

I don't think there is an official way to have your own branch, or to change the path. There are a few ways to do it. You can always just checkout your own branch in `../modules/lang/rust`, but be aware that doing `west update` will switch that back to the one in your modules file.

The other way is to edit the `zephyr/submanifests/optional.yaml` to point to yours. You could just change the path, or add a whole new repo. But, it isn't very set up to do local development on the modules.

I just do my work directly in `../modules/lang/rust`.

Issue #113, opened 2025-07-01
`labels` submodule is duplicated everywhere in devicetree.


d3zd3z:
The generated devicetree is supposed to have a `labels` node at the top level. However, it seems that it is generated within every node in the generated devicetree. This makes for weird possible paths, as well, as makes the generated devicetree file significantly larger than it needs to be.

PR #114, opened 2025-07-01
dt-rust: Fix "Root" node detection

d3zd3z:
The labels node should occur only at the root node in the Rust-converted devicetree.  However, when translating the format of the yaml file to use explicit types, the rule for this node didn't end up in an array. It isn't clear why this doesn't generate an error, but making it an explicit array does fix the issue.

Without this fix, the devicetree.rs that is generated has a duplicate copy of the labels module under every node.  This fixes it to only occur at the top level.

Fixes: #113

### DIFF
--- a/dt-rust.yaml
+++ b/dt-rust.yaml
@@ -66,6 +66,7 @@
 
 # Generate a pseudo node that matches all of the labels across the tree with their nodes.
 - name: labels
-  rules: !Root
+  rules:
+    - !Root
   actions:
     - !Labels


teburd (approved):

teburd:
Might need a little helper if you using serde and you want non-empty arrays https://stackoverflow.com/questions/75016452/serde-how-to-error-out-when-deserializing-empty-arrays

yaml is weird and issues like this might just be ignored :shrug: That's not unusual for yaml stuff.

d3zd3z:
> Might need a little helper if you using serde and you want an empty array to error out https://stackoverflow.com/questions/75016452/serde-how-to-error-out-when-deserializing-empty-arrays
> 
> yaml is weird and issues like this might just be ignored 🤷 That's not unusual for yaml stuff.

The thing is, I'm not sure an empty set of rules is exactly wrong. Well, maybe it is, as it would mean the actions should apply to every node. But, maybe there will be a future action where that makes sense. I'm not quite sure that the `!Root` means when just the value of the `rules:`, but clearly it does come out as an empty array.

cfriedt (approved):

PR #127, opened 2025-08-19
Check in Cargo.lock files

d3zd3z:
A recent short-term broken release of a transient dependency broke CI builds of Zephyr with Rust support. Although it was fixed shortly thereafter, it is a problem we'd like to avoid.

Until now, we've avoided committing the lock files.  This change commits the lock files for each sample and test, building them with versions that have been tested.


kurtjd (approved):
Looks good, though will likely need to be updated after rebasing to account for new samples.

felipebalbi (approved):

PR #128, opened 2025-08-22
Add support for nrf54l15dk/nrf54l15/cpuapp

masz-nordic:


eHammarstrom (approved):

d3zd3z (commented):
Sorry for taking so long to get around to this. As you can see, it does need to be rebased onto the latest version, but otherwise looks good.

felipebalbi (approved):
straightforward

PR #129, opened 2025-08-22
Add support for nrf54l15dk/nrf54l15/cpuflpr

masz-nordic:
Depends on:
https://github.com/zephyrproject-rtos/zephyr-lang-rust/pull/128
https://github.com/zephyrproject-rtos/zephyr-lang-rust/pull/108


samples/hello_world/src/lib.rs
```diff
@@ -11,5 +11,5 @@ extern "C" fn rust_main() {
         zephyr::set_logger().unwrap();
     }
 
-    info!("Hello world from Rust on {}", zephyr::kconfig::CONFIG_BOARD);
+    info!("Hello world from Rust on {}", zephyr::kconfig::CONFIG_BOARD_TARGET);
```
d3zd3z: You'll need to either format these manual, or just use `cargo fmt` to reflow them. With the longer name, it breaks this across multiple lines.

d3zd3z (approved):
Looks good, and thank you for this contribution. This should help us get more targets that don't use build configurations that exactly match one tier 1 or 2 of Rust.

I think there is still a rustfmt issue with one of the longer names, but I'm not seeing it locally.

eHammarstrom (approved):

PR #132, opened 2025-09-21
samples: rust: blinky: Use conditional imports to fix compiler warnings

tinegachris:
## Summary

This PR fixes compiler warnings in the Rust blinky sample by moving imports inside the conditional compilation blocks where they are actually used.

## Problem

When building the blinky sample for targets without an `led0` device tree alias (like `qemu_riscv32`), the compiler generates warnings about unused imports:

## Solution

Move the imports (`ZR_GPIO_OUTPUT_ACTIVE`, `sleep`, `Duration`) inside the `#[cfg(dt = "aliases::led0")]` conditional block where they are actually used.

## Changes Made

- Moved `use zephyr::raw::ZR_GPIO_OUTPUT_ACTIVE;` inside `#[cfg(dt = "aliases::led0")]` block
- Moved `use zephyr::time::{sleep, Duration};` inside `#[cfg(dt = "aliases::led0")]` block
- No functional changes to the logic

### DIFF
--- a/samples/blinky/src/lib.rs
+++ b/samples/blinky/src/lib.rs
@@ -10,9 +10,6 @@
 
 use log::warn;
 
-use zephyr::raw::ZR_GPIO_OUTPUT_ACTIVE;
-use zephyr::time::{sleep, Duration};
-
 #[no_mangle]
 extern "C" fn rust_main() {
     unsafe {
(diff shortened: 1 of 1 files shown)


d3zd3z (approved):

d3zd3z (changes requested):
Ahh, I missed that the PR needs a signed-off-by in the commit. This should be checked by CI, and I'm investigating why the workflows didn't run for this change.

tinegachris:
@d3zd3z Thanks! I have added the Signed-off-by and force-push. Please let me know what you discover about the workflows.

eHammarstrom (approved):

d3zd3z (approved):

PR #133, opened 2025-10-13
fix(zephyr-sys): bindgen stddef.h not found error

ttwards:
Fix bindgen build failure caused by missing system headers by:

- Pass CC environment variable from CMake to cargo build/doc commands
- Dynamically detect GCC toolchain path and version from CC variable
- Add GCC include directory to clang arguments for bindgen
- Create dummy arm_acle.h header to avoid ARM ACLE parsing errors

The root cause was that bindgen couldn't find stddef.h and other GCC builtin headers because:
1. CC variable was not passed to cargo, preventing toolchain detection
2. GCC include path was not provided to clang
3. ARM ACLE headers caused constant expression errors in bindgen

Tested on: qemu_cortex_m3, Zephyr SDK 0.17.2

### DIFF
--- a/zephyr-sys/build.rs
+++ b/zephyr-sys/build.rs
@@ -53,8 +53,31 @@ fn main() -> Result<()> {
     let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
     let wrapper_path = PathBuf::from(env::var("WRAPPER_FILE").unwrap());
 
+    // Get GCC toolchain info to find stddef.h
+    let cc = env::var("CC").unwrap_or_default();
+    let gcc_include = if !cc.is_empty() {
+        let cc_path = Path::new(&cc);
+        let toolchain_path = cc_path.parent().and_then(|p| p.parent()).unwrap();
+        let target_triple = cc_path.file_name().unwrap().to_str().unwrap()
+            .strip_suffix("-gcc").unwrap_or("arm-zephyr-eabi");
+        let gcc_version = std::process::Command::new(&cc)
+            .arg("-dumpversion")
+            .output().ok()
+            .and_then(|o| String::from_utf8(o.stdout).ok())
+            .map(|v| v.trim().to_string())
+            .unwrap_or("12.2.0".to_string());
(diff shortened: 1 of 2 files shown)


zephyr-sys/build.rs
```diff
@@ -53,8 +53,31 @@ fn main() -> Result<()> {
...
+        let cc_path = Path::new(&cc);
+        let toolchain_path = cc_path.parent().and_then(|p| p.parent()).unwrap();
+        let target_triple = cc_path.file_name().unwrap().to_str().unwrap()
+            .strip_suffix("-gcc").unwrap_or("arm-zephyr-eabi");
+        let gcc_version = std::process::Command::new(&cc)
+            .arg("-dumpversion")
+            .output().ok()
+            .and_then(|o| String::from_utf8(o.stdout).ok())
+            .map(|v| v.trim().to_string())
+            .unwrap_or("12.2.0".to_string());
```
eHammarstrom, line 68: Why are we defaulting to 12.2.0?

zephyr-sys/build.rs
```diff
@@ -53,8 +53,31 @@ fn main() -> Result<()> {
     let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
     let wrapper_path = PathBuf::from(env::var("WRAPPER_FILE").unwrap());
 
+    // Get GCC toolchain info to find stddef.h
+    let cc = env::var("CC").unwrap_or_default();
```
d3zd3z, line 57: So, I don't think it is the right thing, _ever_ to use the gcc compiler-specific header files with the clang compiler. Clang will have it's own version of these headers. Some platforms will have clang try to figure this out, but that doesn't happen everywhere, and I'm guessing you're running into that.

It's a little weird to try to figure out the clang headers, but it ends up being something like:
```
RESOURCE_DIR=$(${LIBCLANG_PATH}/../bin/clang --print-resource-dir)
```
Then the arguments to the clang invocation need to include: `nostdinc` (which I think must already be the case), and then `-isystem $RESOURCE_DIR/include`. This should get the versions of these headers that make sense with the clang compiler.

d3zd3z (changes requested):

PR #137, opened 2025-12-16
Add gpio keys button sample compatible with SW0 and LED0 boards

eHammarstrom:
A lot of devkits support the SW0 and (optional) LED0 aliases, as such having a sample to run (or get started with) on these boards may serve as a good starting point when bringing up other boards.


samples/button/sample.yaml
```diff
@@ -0,0 +1,21 @@
...
+    # - qemu_riscv64
+    # - nrf52840dk/nrf52840
+    - nrf5340dk/nrf5340
+tests:
+  sample.rust.basic.button:
+    tags:
+      - button
+      - gpio
+    filter:
+      - dt_enabled_alias_with_parent_compat("sw0", "gpio-keys")
```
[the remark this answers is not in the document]
eHammarstrom: Invalid comment due to the previous comment about `led0` being optional in the code, which it is. README has been updated to reflect this.

samples/button/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+
+#[embassy_executor::task]
+async fn button_listener(mut button: GpioPin) -> ! {
```
d3zd3z, line 87: As an example, this is probably fine. If we wanted a better example that included debouncing, that could be added, but I don't see a reason to do that.

samples/button/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+    let sender = IO_CHANNEL.sender();
+    loop {
+        unsafe {
```
d3zd3z, line 90: Unsafe blocks should have // SAFETY:: comments. But, I won't block this, because I don't have clippy running as part of CI yet.

d3zd3z (approved):

samples/button/src/lib.rs
eHammarstrom: @d3zd3z  I'm not well versed in Clippy & github workflows, but would something like this be desirable? https://github.com/zephyrproject-rtos/zephyr-lang-rust/compare/main...eHammarstrom:zephyr-lang-rust:cmake/clippy-linting

PR #148, opened 2026-04-09
Add PWM LED device bindings

d3zd3z:
## Summary
- Add Zephyr LED bindings and Rust LED wrapper for pwm-leds devices
- Extend devicetree augmentation to expose pwm-led controllers and child LED instances
- Include Tiny 2040 hardware test that drives three LEDs through the LED API

## Test plan
- [x] Build and run tests/drivers/pwm-led on tiny2040


dt-rust.yaml
```diff
@@ -42,6 +42,33 @@
...
+- name: pwm-led-controller
+  rules:
+    - !Compatible
+      names: [pwm-leds]
+      level: 0
+  actions:
+    - !Instance
+      raw: !Myself
+      device: crate::device::led::LedController
+
```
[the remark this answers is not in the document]
d3zd3z: Addressed

zephyr/src/device/led.rs
```diff
@@ -0,0 +1,95 @@
...
+///
+/// This wrapper maps to Zephyr's `led_*` API. Devicetree child nodes of a `pwm-leds` controller
+/// become instances of this type.
+#[allow(dead_code)]
+pub struct Led {
+    pub(crate) device: *const raw::device,
+    pub(crate) index: u32,
+}
+
+impl core::fmt::Debug for Led {
```
kurtjd: Would just a `#derive(Debug)` work here? This manual implementation does not seem to differ much from what the derived implementation does.

zephyr/src/device/led.rs
```diff
@@ -0,0 +1,95 @@
...
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("Led")
+            .field("device", &self.device)
+            .field("index", &self.index)
+            .finish()
+    }
+}
+
+// SAFETY: `Led` holds a raw device pointer and an LED index with no thread-affine Rust state.
+unsafe impl Send for Led {}
```
kurtjd: From the Rust wrapper perspective this seems safe, but from my current understanding, all `Led` instances ultimately share the same LED controller driver. Can we guarantee that simultaneous accesses to the back-end C driver won't cause data races?

zephyr/src/device/led.rs
```diff
@@ -0,0 +1,95 @@
...
+    ) -> Option<LedController> {
+        if !unique.once() {
+            return None;
+        }
+
+        Some(LedController { device })
+    }
+
+    /// Verify that the underlying LED controller is ready for use.
+    pub fn is_ready(&self) -> bool {
```
kurtjd: Noticed the pwm led test doesn't really make use of this. Do we ultimately just need `LedController` as a stub as part of augment.rs/devicetree.rs build process? What I mean is it seems `LedController` isn't really meant to be used publicly, instead most application logic going through individual `Led` instances instead.

zephyr/src/device/led.rs
```diff
@@ -0,0 +1,95 @@
...
+        unsafe { raw::device_is_ready(self.device) }
+    }
+}
+
+/// A single logical LED exposed by a Zephyr LED controller.
+///
+/// This wrapper maps to Zephyr's `led_*` API. Devicetree child nodes of a `pwm-leds` controller
```
kurtjd: Just confirming, ultimately this wrapper should be generic for any LED driver (not just the `pwm-leds` driver), but right now we only can generate this specifically for pwm-leds right (since there isn't really a rule to match on for all LED drivers I guess).

kurtjd (approved):
Looks good! Mostly just a few non-blocking questions really to help my understanding, so feel free to ignore if you've already considered them.

zephyr/src/device/led.rs
d3zd3z: I didn't actually try it. I'll see if that works.

d3zd3z: Tried it, and yes — `#[derive(Debug)]` works and produces identical output, since raw pointers implement `Debug` unconditionally so there's no bound to add. Builds and Clippy clean on hifive1. Changed.

zephyr/src/device/led.rs
d3zd3z: We can't guarantee it, and it isn't really specific to LEDs — the same question applies to GPIO and to every device wrapper here.

Concretely for this path: `led_pwm_set_brightness()` only reads `dev->config` (a const per-LED `pwm_dt_spec`) and forwards to `pwm_set_pulse_dt()`, so the LED driver adds no shared mutable state of its own. One layer down, neither `pwm_sifive.c` nor `pwm_rpi_pico.c` takes a lock, and on RP2040 the A/B channels of a slice share registers — so two threads driving sibling LEDs on one slice can race there. `led.h` documents no guarantee either way.

The general problem is that "can this device be shared" is a per-driver property, and Zephyr doesn't expose it in any form we could map onto the device structure, so there's nothing for the binding layer to key off. Some controllers are entirely safe to share; others are not. Meanwhile `Send` isn't really optional: without it a device is close to useless in a multithreaded application. `Sync` is the one being withheld, so at least a single `Led` can't be shared by reference across threads.

I don't have a good answer for this yet, and I'd rather not have the wrapper imply it has one. It's worth treating as a cross-cutting design question for the bindings rather than settling it here.

zephyr/src/device/led.rs
d3zd3z: Yes, essentially. The `!Instance` augmentation emits `get_instance_raw()`, `get_static_raw()`, `UNIQUE`, `STATIC` and the typed `get_instance()` as a single unit (`zephyr-build/src/devicetree/augment.rs`), and the generated children do `let device = super::get_instance_raw();`. So the parent node needs that augmentation to exist, and `LedController` is the typed by-product that comes with it — in the generated devicetree for this test it appears only in its own getter, and nothing else refers to it.

I wouldn't lean on `is_ready()` as a justification either; I'm not convinced we should be exposing it per-object at all. It answers "has this device been initialized", intended for pointers captured from `DEVICE_DT_GET()`, and arguably belongs in the constructor rather than on every wrapper.

Dropping the type would mean teaching the augmentation machinery a raw-only instance action that emits the raw accessors without a wrapper type. Worth doing, but a codegen change rather than something for this PR.

zephyr/src/device/led.rs
d3zd3z: Right on both counts. The wrapper only calls the generic `led_*` API, so it is already driver-agnostic; the restriction lives in `dt-rust.yaml`, which matches `names: [pwm-leds]`, and extending it is a matter of listing more compatibles.

One wrinkle specifically for `gpio-leds`: there's already a `gpio-leds` rule that turns each child into a `GpioPin` via its `gpios` phandle, so adding that compatible to the LED rule would have the same child nodes matched by two augmentations. Other LED controllers (the I2C ones, say) should be a straight addition.

PR #150, opened 2026-04-09
Add I2C controller and target device bindings

d3zd3z:
## Summary
- Add Rust wrapper for Zephyr's blocking I2C controller API (write, read, write_read)
- Add I2C target registration API (register/unregister with extern "C" callbacks)
- Controller test exercising plain write/read and combined write_read transactions
- Target test implementing a 16-byte register file at address 0x42
- Fix shared target state to use SpinMutex instead of unsafe raw pointer aliasing

**Note:** Based on #146 (davidb-dt-all-properties). Will retarget to main after that merges.

## Test plan
- [ ] Build and run tests/drivers/i2c-controller on tiny2040
- [ ] Build and run tests/drivers/i2c-target on tiny2040
- [ ] Pair controller + target tests on two boards


zephyr-sys/build.rs
```diff
@@ -120,6 +120,12 @@ fn main() -> anyhow::Result<()> {
         // LED Strip
         .allowlist_item_if("led_rgb", || options.contains("CONFIG_LED_STRIP"))
         .allowlist_function_if("led_strip_.*", || options.contains("CONFIG_LED_STRIP"))
+        // I2C
+        .allowlist_function_if("i2c_transfer", || options.contains("CONFIG_I2C"))
```
kurtjd: Can we do something like `i2c_.*` here so we don't need the manual shims below in `wrapper.h`?

zephyr-sys/wrapper.h
```diff
@@ -143,3 +144,46 @@ static inline int zr_irq_lock(void) {
...
+ */
+const uint8_t ZR_I2C_MSG_WRITE = I2C_MSG_WRITE;
+const uint8_t ZR_I2C_MSG_READ = I2C_MSG_READ;
+const uint8_t ZR_I2C_MSG_STOP = I2C_MSG_STOP;
+const uint8_t ZR_I2C_MSG_RESTART = I2C_MSG_RESTART;
+const uint8_t ZR_I2C_MSG_ADDR_10_BITS = I2C_MSG_ADDR_10_BITS;
+
+/*
+ * I2C convenience functions are static inline in Zephyr, so we need shim
+ * wrappers for bindgen to pick them up.
```
kurtjd: As mentioned above, are these needed? It looks like in build.rs you already have `wrap_static_fns(true)` so it seems bindgen should automatically pick them up.

zephyr/src/device/i2c.rs
```diff
@@ -0,0 +1,430 @@
...
+// Re-export the raw callback types so users can write `extern "C"` handlers without reaching into
+// `zephyr::raw` directly.
+pub use raw::{i2c_target_callbacks, i2c_target_config};
+
+/// A single operation in a multi-stage I2C transaction.
+///
+/// Used with [`I2c::transfer`] to compose general scatter/gather transfers.  A RESTART is
+/// inserted automatically between adjacent operations of differing direction; the underlying
+/// `i2c_transfer` always emits a STOP after the final operation.
+pub enum Operation<'a> {
```
kurtjd: I would recommend using the existing [embedded_hal::i2c::Operation](https://docs.rs/embedded-hal/1.0.0/embedded_hal/i2c/enum.Operation.html) enum here which is identical. Also think now would be a good time to implement the [embedded_hal::i2c::I2c](https://docs.rs/embedded-hal/1.0.0/embedded_hal/i2c/trait.I2c.html) trait.

zephyr/src/device/i2c.rs
```diff
@@ -0,0 +1,430 @@
...
+    ///
+    /// Wraps Zephyr's `i2c_transfer`, the general-purpose scatter/gather entry point.  Each
+    /// [`Operation`] becomes one `i2c_msg`; a RESTART is inserted between adjacent operations
+    /// whose direction differs, and `i2c_transfer` itself appends a STOP after the final
+    /// message.
+    ///
+    /// Up to [`MAX_TRANSFER_OPS`](Self::MAX_TRANSFER_OPS) operations are supported per call;
+    /// passing more returns `Err(Error(E2BIG))`.  For simple cases prefer
+    /// [`write`](Self::write), [`read`](Self::read), or [`write_read`](Self::write_read).
+    pub fn transfer(&mut self, addr: u16, ops: &mut [Operation<'_>]) -> Result<()> {
```
kurtjd: Instead of hardcoding a max limit, could you make this generic over a const N, as in something like:
`pub fn transfer<const N: usize>(&mut self, addr: u16, ops: &mut [Operation<'_>]; N) -> Result<()>`

Then you could likely factor out most of the body into a [non-generic inner function](https://www.possiblerust.com/pattern/non-generic-inner-functions) to prevent monomorphization bloat.

zephyr/src/device/i2c.rs
```diff
@@ -0,0 +1,430 @@
...
+        to_result_void(raw::zr_i2c_target_register(self.device, config_ptr))?;
+        Ok(I2cTarget {
+            device: self.device,
+            config: config_ptr,
+        })
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Safe I2C target abstraction
```
kurtjd: Haven't looked over the i2c target implementation here in too much detail since the zephyr approach to i2c target seems different than what I'm used to so will need to spend a bit more time familiarizing myself with it. But I would recommend having an i2c parent module and splitting i2c controller and i2c target wrappers into separate Rust files.

kurtjd (approved):
Overall looks good and ready for merge, just a few minor suggestions that might be worth considering.

zephyr-sys/build.rs
d3zd3z: Done — the allowlist is now `i2c_.*`, matching how `uart_.*` and `led_.*` are already handled. You were right that this makes the shims unnecessary: they only existed because the allowlist named `i2c_transfer` alone, so `wrap_static_fns` never saw the rest.

zephyr-sys/wrapper.h
d3zd3z: Confirmed and removed. With the allowlist broadened, bindgen generates `i2c_write`, `i2c_read`, `i2c_write_read`, `i2c_target_register` and `i2c_target_unregister` itself, and emits the C side into the generated `wrapper.c` — I checked the generated bindings and wrapper for all five before deleting the shims. The `ZR_I2C_MSG_*` constants stay, since those wrap macros rather than static inline functions.

zephyr/src/device/i2c.rs
d3zd3z: There's real interest in implementing the embedded-hal traits, so I don't want to dismiss this — but I'd rather settle it as its own discussion than fold it into this PR. embedded-hal isn't currently a dependency anywhere in the crate, so adopting it here sets the direction for every other wrapper (GPIO first among them) and pins a public API version. There's also the question of the HALs that don't map cleanly onto Zephyr — storage being the obvious one — which argues for deciding where the boundary is before starting rather than after.

Keeping the local `Operation` for now, and leaving this open.

zephyr/src/device/i2c.rs
d3zd3z: Agreed with your edit — the `embedded_hal::i2c::I2c::transaction` signature takes a non-generic `&mut [Operation]`, so going const-generic here would have to be undone the moment we implement the trait. Leaving the hardcoded `MAX_TRANSFER_OPS` with the `E2BIG` error for now; it's the honest no-alloc tradeoff, and it keeps the door open for the trait.

zephyr/src/device/i2c.rs
d3zd3z: Done — `device::i2c` is now a parent module re-exporting the public types, with the implementation in `device/i2c/controller.rs` (160 lines) and `device/i2c/target.rs` (271 lines). The external API is unchanged. `I2cTarget`'s fields became `pub(super)` so that the controller's `register_target()` can still construct it from the sibling module.

No rush on the target review — the Zephyr target model is genuinely different from most, and the split should at least make it easier to read in isolation.

PR #152, opened 2026-04-15
ci: Added macOS runner

Dooleweerdt:
Added build workflow running on a macOS runner.
Added env variables to use llvm/clang from homebrew paths. Disable Renode install on the macOS runner (unused?) Moved linux specific commands to a linux specific step.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-22.04]
+        os: [ubuntu-22.04, macos-15]
     runs-on: ${{ matrix.os }}
     env:
       WGET_ARGS: "-q"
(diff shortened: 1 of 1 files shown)


Dooleweerdt:
Hi @d3zd3z 
I was playing around with adding the macOS runner on my own project, so why not add it to the zephyr-lang-rust repo builds too? ;-)
What do you think?

.github/workflows/build.yml
```diff
@@ -59,6 +75,7 @@ jobs:
 
       # Note that Renode is only provided for x86_64 targets.  This matches the github runners.
       - name: Install Renode
+        if: runner.os == 'ubuntu'
```
d3zd3z: This is never true. I believe on ubuntu, the runner.os will be `'Linux'`.

.github/workflows/build.yml
```diff
@@ -45,6 +45,22 @@ jobs:
           manifest-file-name: ci-manifest.yml
           toolchains: arm-zephyr-eabi:riscv64-zephyr-elf
 
+      - name: Setup macOS LLVM Environment
+        if: runner.os == 'macOS'
+        run: |
+          # Set hardcoded paths for LLVM 18 installed via Homebrew on macOS 15 - fits with GitHub's macOS 15 image
+          echo "LIBCLANG_PATH=/opt/homebrew/Cellar/llvm@18/18.1.8/lib/libclang.dylib" >> $GITHUB_ENV
```
d3zd3z: Perhaps something like:
```bash
LLVM_PREFIX=/opt/homebrew/opt/llvm/lib
RESOURCE_DIR="$("$LLVM_PREFIX/bin/clang" -print-resource-dir)"
export BINDGEN_EXTRA_CLANG_ARGS="-resource-dir=$RESOURCE_DIR"
```
this should keep it a little less fragile as to specific version.
Any reason to not use macos-26? I build this regularly on Tahoe. And, I haven't needed the int64 override definition below.

d3zd3z (changes requested):

.github/workflows/build.yml
Dooleweerdt: Yes it is Linux. Sorry I didn't catch it myself.
@d3zd3z:Do you know why Renode is added to the workflow? As far as I can see it is not used? We could clean it up?

.github/workflows/build.yml
Dooleweerdt: I have now changed it to use macos-26. I cleaned out the int64 override definitions - I apparently need them in my own projects workflow, but they are not needed here.

.github/workflows/build.yml
d3zd3z: I suspect I just copied that from an existing workflow. I think there might have been some goal of renode, but it isn't part of the SDK, just installed in the docker image that Zephyr main uses, which is too large for the free runner github provides.

Dooleweerdt: @d3zd3z: I can clean it up, if it's ok with you?
I can also leave it in...

PR #156, opened 2026-05-01
Migrate paste to pastey (RUSTSEC-2024-0436)

mbUSC:
The paste crate has been archived by its maintainer (see the announcement on GitHub) and now carries RUSTSEC-2024-0436 "no longer maintained." Downstream consumers running cargo-deny / cargo-audit either fail their advisory checks or have to add an ignore entry per project.                                   
                                                                  
pastey is the maintained fork and is documented as a drop-in replacement: same `paste!` macro, same grammar, same behavior. Upstream usage in zephyr-lang-rust is limited to one re-export (`pub use paste::paste` in `zephyr/src/lib.rs`) and two callsites inside macro definitions in `zephyr/src/object.rs` (`_kobj_stack` macro arms). Because the callsites use `$crate::paste! { ... }`, only the re-export needs to change.

### DIFF
--- a/zephyr-build/Cargo.toml
+++ b/zephyr-build/Cargo.toml
@@ -5,6 +5,7 @@
 name = "zephyr-build"
 version = "0.1.0"
 edition = "2021"
+license = "Apache-2.0"
 description = """
 Build-time support for Rust-based applications that run on Zephyr.
 Provides utilities for accessing Kconfig and devicetree information.
--- a/zephyr-sys/Cargo.toml
+++ b/zephyr-sys/Cargo.toml
@@ -5,6 +5,7 @@
 name = "zephyr-sys"
 version = "0.1.0"
 edition = "2021"
+license = "Apache-2.0"
 description = """
 Zephyr low-level API bindings.
 """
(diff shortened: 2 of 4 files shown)


d3zd3z (changes requested):
Thank you for your contribution. The change itself looks good. However, the commit text needs a little bit of work.

As this is part of the Zephyr project, commits should follow the [Contribution Guidelines](https://docs.zephyrproject.org/latest/contribute/guidelines.html) from Zephyr, in this case, specifically, each commit should follow the "Commit Message Guidelines", including having a summary line (area might just be "zephyr:" for the paste change, and perhaps "cargo:" or something like that for the license change). The description should include what the change does, why it was done. It doesn't need to be long, but there commit body can't be empty.

And, then importantly, each commit must have a Signed-off-by line, indicating compliance with the [Developer's Certificate of Origin](https://docs.zephyrproject.org/latest/contribute/guidelines.html#developer-certification-of-origin-dco)

felipebalbi (approved):
Once @d3zd3z concerns about commit log quality are sorted out, this should be good to go.

PR #163, opened 2026-06-17
Add Rust port of BMI270 6-axis IMU sensor sample

yuvraj-cirrent:
Add Rust port of BMI270 6-axis IMU sensor sample specifically configured for Infineon cy8ckit_062s2_ai board. The sample uses the accel0 alias pointing to BMI270 on I2C at address 0x68.

This is a port of bmi270 sensor sample example in C to Rust.
https://github.com/yuvraj-cirrent/zephyr/tree/main/samples/sensor/bmi270


yuvraj-cirrent:
@d3zd3z Can you please approve workflow for this PR?

samples/sensor/bmi270/src/sensor_shim.c
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright (c) 2021 Bosch Sensortec GmbH
+ * Copyright (c) 2025 (Rust port shim)
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Thin C bridge so Rust can reach Zephyr's sensor API.
+ *
+ * Why this file exists: sensor_sample_fetch(), sensor_channel_get(),
```
kurtjd: Can you not already access these with `zephyr::raw` and `zephyr::time`?

kurtjd: Ah, I see. These aren't already part of the allowlist, but your other PR added them to the allowlist to auto generate the wrappers instead of needing these manual shims.

I am curious what the best approach is for this. Because I've seen, for example, in the i2c PR open (https://github.com/zephyrproject-rtos/zephyr-lang-rust/pull/150) the author takes the manual shim approach as well, but I suspect just adding the functions to the allowlist would work too.

yuvraj-cirrent: @kurtjd Removed C shim layer. Now using node from wrapper.h

felipebalbi (approved):

PR #164, opened 2026-06-18
Add pressure polling sensor sample in Rust for DPS368

yuvraj-cirrent:
Add Rust port of pressure_polling sample specifically configured for Infineon cy8ckit_062s2_ai board. The sample uses the pressure-sensor alias pointing to dps368 on I2C.

This is a port of dps368 sensor sample example in C to Rust.
https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/sensor/pressure_polling


yuvraj-cirrent:
@d3zd3z Can you please approve workflow for this PR?

samples/sensor/pressure_polling/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    do_polling();
+}
+
+#[cfg(dt = "aliases::pressure_sensor")]
+fn do_polling() {
+    // SAFETY: We access only static device structures defined by the Zephyr device tree,
+    // and call only Zephyr C sensor API functions that are safe to call from a single thread
+    // after device initialization.
+    unsafe {
```
kurtjd: I would recommend keeping the `unsafe { ... }` blocks limited in scope to where necessary (similar to your other example: https://github.com/zephyrproject-rtos/zephyr-lang-rust/pull/163) instead of one massive unsafe block. Would be helpful for demonstration purposes to explicitly show where unsafe is actually required, and the additional safety comment per unsafe block would be helpful to show why the way it is being used is safe.

yuvraj-cirrent: @kurtjd Addressed above comment

kurtjd:
Can wait for others' opinions as well, but I think it makes sense to condense this and your other PR (https://github.com/zephyrproject-rtos/zephyr-lang-rust/pull/163) into a single sensor sample that tries to be as generic as possible in just showing how to interface with the raw sensor API from Rust. I don't think we necessarily need two different samples just yet showing how to interface with specific devices.

Ideally in the future (which I'm hoping to get to soon) we have a safe sensor wrapper similar to gpio/flash, but that is beyond the scope of these PRs presumably. There is still value in showing the raw sensor API accesses in application code like you are doing, I just think you should keep it to one sample and try to keep it as device agnostic as possible.

Arguably there could be value in a sample for each **type** of sensor, but that might be more useful once the safe Rust wrapper is introduced. But for now, with the raw sensor API calls which all look similar regardless of type of sensor, I think we should keep it to one sample.

PR #167, opened 2026-07-16
Add Fuel Gauge driver to zephyr::device

bjackson312006:
This PR adds a `zephyr::device::fuel_gauge` module (located at `zephyr/src/fuel_gauge.rs`) that implements a safe Rust wrapper around the [Zephyr fuel gauge API](https://docs.zephyrproject.org/latest/hardware/peripherals/fuel_gauge.html). This PR also adds a simple `fuel_gauge` sample to demonstrate this driver.

Since Zephyr's fuel gauge API is based around the `fuel_gauge_prop_val` union, this driver was designed to expose all the same data but via explicit accessors for each field/property. The C `fuel_gauge_prop_t` type was adapted into `pub(crate) enum FuelGaugeProp` and `pub(crate) enum FuelGaugeBufferProp`, which are meant to provide some internal type safety for the rest of the driver. These enums map directly to the autogenerated `crate::raw::fuel_gauge_prop_type` module. This involves some annoying boilerplate, but it's worth it so the raw C constants can be contained in one place without needing to be used internally by the rest of the driver. Plus, these enums will never need changes as long as Zephyr's API stays stable. 
^^ Note: The depreciated constants in `crate::raw::fuel_gauge_prop_type` were deliberately left out of these enums.

Below this, there are some internal helpers used to wrap around the raw Zephyr calls: `pub(crate) fn get_prop()`, `pub(crate) fn set_prop()`, and `pub(crate) fn get_buffer_prop()`. These helpers are meant to isolate the main `unsafe` blocks in the driver, and allow the rest of the driver to use the safe `FuelGaugeProp` and `FuelGaugeBufferProp` types when requesting data. TLDR they provide safer internal wrappers for the raw `fuel_gauge_get_prop()`, `fuel_gauge_set_prop()`, and `fuel_gauge_get_buffer_prop()` calls to the C API.

After that, there's the `ManufacturerName`, `DeviceName`, and `DeviceChemistry` types, which each wrap around a `heapless::String`. This adds a dependency (`heapless`), but significantly improves the interface of the user-facing `manufacturer_name()`, `device_name()`, and `device_chemistry()` functions since users no longer need to manually pass in a mut buffer for the calls to write into. Plus, a `String` is a more accurate way to type this data (compared to `&mut [u8]`) and works better with Debug/Display/fmt.

Below that is the main public API, which first includes `is_ready()` and the aforementioned `manufacturer_name()`, `device_name()`, and `device_chemistry()` functions used to read buffer props. The functions after that, which make up the vast majority of the public API, are the getters and setters for the single-value union properties. These functions all follow an identical structure: They call the safe `get_prop()`/`set_prop()` helpers for their respective `FuelGaugeProp`, and then access the returned union prop in a small unsafe block. I looked into moving these unsafe blocks directly into `get_prop()` and `set_prop()`, but this wasn't possible to do without defeating the purpose of the moves in the first place. However, since these blocks are all tiny, identical, and performing an action we know to be safe (union operations guaranteed to valid per the calls to the Zephyr API), I think this is fine.
^^ Note: Not all of the getters have setters since a lot of the fields are only meant to be read

There's no single property to match on for a fuel gauge as far as I know, so I'm just matching on a single property (`"sbs,sbs-gauge-new-api"`) like flash-controller does. This should be fine for SBS-compatible fuel gauges, but down the line it may be necessary to explore some other way to automatically make this work for all fuel gauges.

Aside that, the new `fuel_gauge` sample shows how to use the driver by periodically reading a bunch of data from the gauge and logging it. The logging is made easy since all types returned by the driver derive `Debug`, including `ManufacturerName`, `DeviceName`, and `DeviceChemistry`.


samples/fuel_gauge/boards/mimxrt685_evk_mimxrt685s_cm33.conf
```diff
@@ -0,0 +1,24 @@
...
+CONFIG_RTT_CONSOLE=y
+CONFIG_LOG_BACKEND_UART=n
+
+# RTT buffer configuration
+CONFIG_SEGGER_RTT_BUFFER_SIZE_UP=4096
+CONFIG_SEGGER_RTT_BUFFER_SIZE_DOWN=2048
+CONFIG_LOG_BUFFER_SIZE=4096
+CONFIG_LOG_MODE_IMMEDIATE=y
+
+# Fuel Gauge
```
kurtjd: Should these belong in `prj.conf` since they seem board agnostic?

samples/fuel_gauge/boards/native_sim_native_64.conf
```diff
@@ -0,0 +1,15 @@
+# Keep Zephyr tickrate consistent with embassy tickrate
+CONFIG_SYS_CLOCK_TICKS_PER_SEC=10000
+
+# Logging
+CONFIG_LOG=y
+CONFIG_LOG_USE_VLA=y
+
+# Fuel Gauge
```
kurtjd: Ditto.

samples/fuel_gauge/src/lib.rs
```diff
@@ -0,0 +1,104 @@
+#![no_std]
```
kurtjd, line 4: Recommend adding the license and copyright header here to align with most of the rest of the source files in this project. I know not all files do currently, but seems to be the most consistent.

zephyr/src/device/fuel_gauge.rs
```diff
@@ -0,0 +1,610 @@
+//! Device wrapper for a fuel gauge.
```
kurtjd, line 4: Ditto about header.

zephyr/src/device/fuel_gauge.rs
```diff
@@ -0,0 +1,610 @@
...
+
+/// A fuel gauge device.
+pub struct FuelGauge {
+    device: *const crate::raw::device,
+}
+
+#[rustfmt::skip]
+#[repr(u32)]
+pub(crate) enum FuelGaugeProp {
+    AvgCurrentUa = crate::raw::fuel_gauge_prop_type::FUEL_GAUGE_AVG_CURRENT_UA,
```
kurtjd: nit: Could be more readable to plop a `use crate::raw::fual_gauge_prop_type` at the top to cutdown on some of the repetition.

zephyr/src/device/fuel_gauge.rs
```diff
@@ -0,0 +1,610 @@
...
+    }
+
+    /// Reads the gauge's `manufacturer_name` into a string.
+    pub fn manufacturer_name(&self) -> crate::error::Result<ManufacturerName> {
+        // Add +1 here since the first byte is the string length
+        let mut buffer: [u8; MANUFACTURER_NAME_STRING_SIZE + 1] =
+            const { [0; MANUFACTURER_NAME_STRING_SIZE + 1] };
+        self.get_buffer_prop(FuelGaugeBufferProp::ManufacturerName, &mut buffer)?;
+
+        let len = buffer[0] as usize;
```
kurtjd: You repeat this string parsing code several times, it would be better to factor it out into a common helper (and you can make the helper generic over N if you want to provide a specific buffer size instead of a max size). You can also make this shorter and more idiomatic (and remove the need for indexing which can introduce possible panic path code if the compiler can't prove the panic will happen).

So something like this:

```rust
fn parse_sbs_str<const N: usize>(buffer: &[u8]) -> crate::error::Result<heapless::String<N>> {
    let (&len, data) = buffer.split_first().ok_or(crate::error::Error(crate::raw::EINVAL))?;
    let data = data.get(..len as usize).ok_or(crate::error::Error(crate::raw::EINVAL))?;
    let s = core::str::from_utf8(data).map_err(|_| crate::error::Error(crate::raw::EILSEQ))?;
    heapless::String::try_from(s).map_err(|_| crate::error::Error(crate::raw::EINVAL))
}
```

And then ManufactuerName can look like:

```rust
    /// Reads the gauge's `manufacturer_name` into a string.
    pub fn manufacturer_name(&self) -> crate::error::Result<ManufacturerName> {
        // Add +1 here since the first byte is the string length
        let mut buffer: [u8; MANUFACTURER_NAME_STRING_SIZE + 1] =
            const { [0; MANUFACTURER_NAME_STRING_SIZE + 1] };
        self.get_buffer_prop(FuelGaugeBufferProp::ManufacturerName, &mut buffer)?;

        Ok(ManufacturerName { inner: parse_str(&buffer)? })
    }
```

zephyr/src/device/fuel_gauge.rs
```diff
@@ -0,0 +1,610 @@
...
+    }
+}
+
+/// Maximum size of the device name string.
+/// According to the Zephyr API docs, device name is 1 byte of string length + 20 bytes of data (21 bytes total). The actual string length is just 20 bytes though.
+const DEVICE_NAME_STRING_SIZE: usize = 20;
+
+/// Represent's a fuel gauge's device name.
+#[derive(Debug)]
+pub struct DeviceName {
```
kurtjd: I think it would make sense just have a single `pub struct SbsString` or something since these all provide the same thing: A wrapper around a `heapless::String`. Unless you imagine situations where we need to differentiate at the type level between a manufacturer name, device name, etc. To avoid making it generic over buffer size you can likely just get way with setting the buffer size to MAX(device name len, manufacturer name len, etc).

This helps avoid code duplication since I recommend implementing a few helper traits/methods here. For example, instead of a parse str helper I mentioned below, you can just implement:

```rust
impl TryFrom<&[u8]> for SbsString {
    type Error = crate::error::Error;

    fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
        <inline parse_str code here>
        Ok(SbsString { inner: s })
    }
}
```

And then you can do for example:
```rust
    /// Reads the gauge's `manufacturer_name` into a string.
    pub fn manufacturer_name(&self) -> crate::error::Result<SbsString> {
        // Add +1 here since the first byte is the string length
        let mut buffer: [u8; MANUFACTURER_NAME_STRING_SIZE + 1] =
            const { [0; MANUFACTURER_NAME_STRING_SIZE + 1] };
        self.get_buffer_prop(FuelGaugeBufferProp::ManufacturerName, &mut buffer)?;

        SbsString::try_from(&buffer[..])
    }
```

Lastly, I recommend implementing `AsRef` here which gives us `as_ref` (basically the same thing as your `as_str` but allows it to be used by any generic code bounded by that trait.

Implementing `Deref` here to get automatic coercion into an &str (e.g. via &SbsString) might be a good idea, or it might fall into the [Deref Polymorphism](https://rust-unofficial.github.io/patterns/anti_patterns/deref.html) antipattern.

Though if you have good reasons to keep it as-is I don't feel strongly.

samples/fuel_gauge/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+/// Struct to store the fuel gauge data we want to get.
+#[allow(dead_code)]
+#[derive(Debug)]
+struct FuelGaugeData {
+    current_ua: i32,
+    voltage_uv: i32,
+    temperature_dk: u16,
+    remaining_capacity_mins: u32,
```
kurtjd: You call this `mins` but `remaining_capacity` returns `uAh`. Is this intentional or are your units off on one end?

samples/fuel_gauge/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+                .inspect_err(|err| log::error!("Failed to read fuel gauge device_name: {}", err))?,
+            device_chemistry: gauge.device_chemistry().inspect_err(|err| {
+                log::error!("Failed to read fuel gauge device_chemistry: {}", err)
+            })?,
+        })
+    }
+}
+
+#[embassy_executor::task]
+async fn fuel_gauge(_spawner: Spawner) {
```
kurtjd: I think you can just drop the spawner parameter here and just not pass it in when you spawn it if you're not using it,

samples/fuel_gauge/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+description = "Sample for fuel gauge driver."
+license = "Apache-2.0 or MIT"
+
+[lib]
+crate-type = ["staticlib"]
+
+[dependencies]
+embassy-executor = { version = "0.7.0", features = ["task-arena-size-4096"] }
+embassy-futures = "0.1.1"
+embassy-sync = "0.8"
```
kurtjd: Are `embassy-futures` and `embassy-sync` used by the sample? If not you can remove these.

zephyr/src/device/fuel_gauge.rs
```diff
@@ -0,0 +1,610 @@
...
+    pub fn as_str(&self) -> &str {
+        self.inner.as_str()
+    }
+}
+
+/// Maximum size of the device chemistry string.
+/// According to the Zephyr API docs, device chemistry is 1 byte of string length + 4 bytes of data (5 bytes total). The actual string length is just 4 bytes though.
+const DEVICE_CHEMISTRY_STRING_SIZE: usize = 4;
+
+/// Represent's a fuel gauge's device chemistry.
```
kurtjd: typo: `Represents`

zephyr/src/device/fuel_gauge.rs
```diff
@@ -0,0 +1,610 @@
...
+    pub fn as_str(&self) -> &str {
+        self.inner.as_str()
+    }
+}
+
+/// Maximum size of the device name string.
+/// According to the Zephyr API docs, device name is 1 byte of string length + 20 bytes of data (21 bytes total). The actual string length is just 20 bytes though.
+const DEVICE_NAME_STRING_SIZE: usize = 20;
+
+/// Represent's a fuel gauge's device name.
```
kurtjd: typo: `Represents`

zephyr/src/device/fuel_gauge.rs
```diff
@@ -0,0 +1,610 @@
...
+            },
+        )
+    }
+}
+
+/// Maximum size of the manufacturer name string.
+/// According to the Zephyr API docs, manufacturer name is 1 byte of string length + 20 bytes of data (21 bytes total). The actual string length is just 20 bytes though.
+const MANUFACTURER_NAME_STRING_SIZE: usize = 20;
+
+/// Represent's a fuel gauge's manufacturer name.
```
kurtjd: typo: `Represents`

samples/fuel_gauge/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.20.0)
+
+list(APPEND DTS_ROOT ${CMAKE_CURRENT_SOURCE_DIR})
```
kurtjd: Is this necessary?

zephyr/src/device/fuel_gauge.rs
```diff
@@ -0,0 +1,610 @@
...
+
+#[repr(u32)]
+pub(crate) enum FuelGaugeBufferProp {
+    ManufacturerName = crate::raw::fuel_gauge_prop_type::FUEL_GAUGE_MANUFACTURER_NAME,
+    DeviceName = crate::raw::fuel_gauge_prop_type::FUEL_GAUGE_DEVICE_NAME,
+    DeviceChemistry = crate::raw::fuel_gauge_prop_type::FUEL_GAUGE_DEVICE_CHEMISTRY,
+}
+
+/// Crate-internal API.
+impl FuelGauge {
```
kurtjd, line 127: If you prefer to have two separate `impl FuelGauge` blocks that's fine, but it feels awkward to place all the helper wrapper types in between them. I would recommend just keeping things logically grouped and organized.

kurtjd (changes requested):
Overall, very good, though a few things I think are worth addressing.

samples/fuel_gauge/boards/mimxrt685_evk_mimxrt685s_cm33.conf
bjackson312006: Even though some of the logging settings could probably be board agnostic, IMO logging as a whole is board-specific. So I kept all the logging settings in the board-specific confs for consistency, since splitting them between each board and the overall .conf seemed like it could get messy if more platforms were to be added

samples/fuel_gauge/boards/native_sim_native_64.conf
bjackson312006: Same as above

samples/fuel_gauge/src/lib.rs
bjackson312006: Added

zephyr/src/device/fuel_gauge.rs
bjackson312006: Added

zephyr/src/device/fuel_gauge.rs
bjackson312006: Fixed

zephyr/src/device/fuel_gauge.rs
bjackson312006: Fixed

zephyr/src/device/fuel_gauge.rs
bjackson312006: Fixed

zephyr/src/device/fuel_gauge.rs
bjackson312006: Fixed

samples/fuel_gauge/CMakeLists.txt
bjackson312006: Nope, removed

zephyr/src/device/fuel_gauge.rs
bjackson312006: I moved the wrapper types up to the top of the file, but kept the two different impl blocks since I feel like it makes sense to have a clear split between the internal helpers and the public API

samples/fuel_gauge/boards/mimxrt685_evk_mimxrt685s_cm33.conf
kurtjd: Sorry I was specifically referring to the settings under `# Fuel Gauge`. I agree though that logging settings are likely board specific though.

samples/fuel_gauge/src/lib.rs
bjackson312006: Fixed, my units were off on the sample end

samples/fuel_gauge/src/lib.rs
bjackson312006: Fixed

samples/fuel_gauge/Cargo.toml
bjackson312006: Nope, removed

zephyr/src/device/fuel_gauge.rs
bjackson312006: I ended going with this via the `FuelGaugeString`. Basically everything as you suggested it, but I kept an explicit as_str() function JIC you know you specifically want a string and don't want to do any manual type annotations if the compiler can't infer the as_ref() type for whatever reason

zephyr/src/device/fuel_gauge.rs
bjackson312006: Basically did this via the try_from() impl from the comment below

samples/fuel_gauge/boards/native_sim_native_64.conf
kurtjd: See: https://github.com/zephyrproject-rtos/zephyr-lang-rust/pull/167#discussion_r3618007495
Let me know if you still don't think they should be in `prj.conf`.

zephyr/src/device/fuel_gauge.rs
kurtjd: Your parsing implementation could still use a bit of cleanup but non-blocking.

samples/fuel_gauge/boards/mimxrt685_evk_mimxrt685s_cm33.conf
bjackson312006: Pushed a new commit to move the `# Fuel Gauge` configs to the sample's main `prj.conf`

samples/fuel_gauge/boards/native_sim_native_64.conf
bjackson312006: Pushed a new commit to move the `# Fuel Gauge` configs to the sample's main `prj.conf`

zephyr/src/device/fuel_gauge.rs
```diff
@@ -0,0 +1,568 @@
...
+    ) -> Option<FuelGauge> {
+        if !unique.once() {
+            None
+        } else {
+            Some(FuelGauge { device })
+        }
+    }
+
+    /// Private helper function to get a fuel gauge prop value.
+    pub(crate) fn get_prop(
```
teburd, line 142: Any reason to make this only pub(crate)?

zephyr/src/device/fuel_gauge.rs
```diff
@@ -0,0 +1,568 @@
...
+            //            This memory area is local to get_prop(), so it will live as long as fuel_gauge_get_prop() is using it.
+            //            In other words, by the time `buffer` goes out of scope, fuel_gauge_get_prop() will have already returned.
+            unsafe { crate::raw::fuel_gauge_get_prop(self.device, prop as u16, &mut buffer) },
+        )?;
+
+        Ok(buffer)
+    }
+
+    /// Private helper function to set a fuel gauge prop value.
+    pub(crate) fn set_prop(
```
teburd, line 161: Like wise, why not expose the underlying function as pub, it is a public API in Zephyr

zephyr/src/device/fuel_gauge.rs
```diff
@@ -0,0 +1,568 @@
...
+                    buffer.as_mut_ptr() as *mut core::ffi::c_void,
+                    buffer.len(),
+                )
+            },
+        )
+    }
+}
+
+/// Public API.
+impl FuelGauge {
```
teburd, line 202: It's nice to have the helpers here actually

zephyr/src/device/fuel_gauge.rs
bjackson312006: The typed accessors for the props are meant to supersede the need to interface with the raw FFI union directly, so making these helpers public would go against the purpose of this driver a bit IMO. I'm not fully against it, but I think if users need to access the raw Zephyr API directly they should just do it explicitly via the `zephyr::raw` functions. Exposing the helpers wouldn't necessarily add any new functionality (since the props are exposed safely via their dedicated getters) and somewhat obfuscates the purpose of this driver in comparison to the existing 1:1 `zephyr::raw` bindings

zephyr/src/device/fuel_gauge.rs
bjackson312006: The true underlying functions that map to the public API in Zephyr are already exposed via `zephyr::raw`, so I think exposing these helpers would be out of place. For accessing the direct public Zephyr API it seems more appropriate to use the existing `zephyr::raw` calls

zephyr/src/device/fuel_gauge.rs
bjackson312006: This probably depends on the two above comments (if the helpers are made public then it makes sense to move them into this impl block, otherwise keeping them in a crate-internals impl block seems to make the most sense)

kurtjd (approved):

PR #169, opened 2026-07-20
zephyr-panic: Move panic handler into crate

kurtjd:
Currently the zephyr crate defines a panic handler implementation, forcing all downstream users into that specific implementation. Ideally users should be able to use the zephyr crate as a library while still having a choice in which panic handler is used. So, the panic handler has been moved into a separate zephyr-panic crate allowing users to opt-in.

One motivation, for example, is the fact that all the metadata associated with panics can significantly bloat binary size. If a user doesn't want to log the panic info or care about it in any way, they can instead choose to use the `panic-halt` crate which allows the compiler to optimize all that out as dead code.

Couple notes:
- I had also considered keeping the panic handler in the `zephyr` crate and making it opt-in via a feature flag, but this just didn't feel right. I think the top-level application code should always explicitly choose a panic handler implementation. This matches the existing panic handling crate ecosystem where you typically see `use panic_halt as _`, `use panic_rtt as _`, etc. Additionally, the feature flag approach leads to the possibility that application code could pull in the zephyr crate transitively from a dep that enables the panic handler, leading to confusion.
- More of a cosmetic issue, but I wanted to name the crate `panic-zephyr` to match the format of other panic crates, but instead chose `zephyr-panic` to keep the name consistent with the other crates in this repo.
- This change was a bit invasive in that I had to modify all the samples/test code (making up the bulk of this PR, although it's just a couple lines changed per sample). This is also a breaking change in that downstream users now need to explicitly opt-in to the panic handler (just like the samples do) while previously it was always included by default.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,7 +131,7 @@ function(rust_cargo_application)
   # For now, hard-code the Zephyr crate directly here.  Once we have
   # more than one crate, these should be added by the modules
   # themselves.
-  set(LIB_RUST_CRATES zephyr zephyr-build zephyr-sys)
+  set(LIB_RUST_CRATES zephyr zephyr-build zephyr-sys zephyr-panic)
 
   get_include_dirs(zephyr_interface include_dirs)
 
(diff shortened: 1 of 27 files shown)


felipebalbi (approved):
I see no issues with this direction. Additionally, the code is quite literally just moving to another location. The added benefit of being able to pull a standard `panic-halt` is also desirable.

d3zd3z (approved):

PR #170, opened 2026-08-02
zephyr: support Rust on runtime-frequency timer targets

gudipudiramanakumar:
## Summary

Enable Rust applications on Zephyr targets that set
`CONFIG_TIMER_READS_ITS_FREQUENCY_AT_RUNTIME`.

The Rust time API now uses tick-backed `Duration` and `Instant` types on
runtime-frequency timer targets. Human-unit conversions delegate to Zephyr's
time conversion helpers, avoiding a compile-time `fugit` frequency requirement.
Compile-time-frequency targets retain the existing `fugit` types.

The bindgen wrapper exposes the Zephyr time-conversion macros required by the
runtime implementation.

## Testing

- Built and linted `samples/hello_world` for
  `xg24_rb4186c/efr32mg24b210f1536im48`.
- Built and linked `samples/blinky` for the same target.
- Flashed and verified `samples/hello_world` on an `xg24_rb4186c` (EFR32MG24B210), observed the Rust hello-world boot message over J-Link CDC UART.


kurtjd (commented):
Is most of this necessary? My understanding is that `CONFIG_SYS_CLOCK_TICKS_PER_SEC` will always be defined even on boards with runtime-frequency support. And the helper macros you wrap use this exact same constant as used here with fugit.

And this is all fugit touches, it only seems to rely on kernel tick rate, NOT cycle rate (which is what's actually guarded by `TIMER_READS_ITS_FREQUENCY_AT_RUNTIME`). So, we shouldn't need a whole new type I don't think, fugit should work fine since `SYS_FREQUENCY` is always static).

Of course, that means we should move the compile error, since really this whole module should work regardless. I think it might have been introduced here because a few samples divide by the `CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC` which will be 0 on runtime-frequency supported boards.

So those samples might need to be reworked instead of just causing an error here.

Anwyay, please let me know if simply removing the compile error allows the current fugit timekeeping to work for your needs.

PR #173, opened 2026-08-12
zephyr: Add `blocking` module with async support

kurtjd:
This adds a new module which makes it very easy to offload blocking Zephyr calls onto a worker thread pool and then `await` them from an async executor thread without stalling the entire executor.

The motivation behind this is that, while many Zephyr subsystems do provide an async interface, some do not (such as fuel gauges), so this allows you to, for example, read from a fuel gauge in an async task without stalling the executor.

Of course, there is always a cost of having to setup a worker pool and offloading work to it in terms of increased memory size and context-switching price, so this is intended to be used simply when you don't have an alternative. But this module handles all the details and provides a simple abstraction making it easy for users who do need this.

Originally had considered using the Zephyr-native P4WQ for the thread pool, but it turns out this won't actually work with blocking calls (see: https://github.com/zephyrproject-rtos/zephyr/pull/92462). Additionally, using a Zephyr-native workqueue doesn't buy us much either since we still need to manage multiple workqueues and the interface for submitting work isn't that helpful here either.

So, in addition to providing the async interface, this also includes simple worker thread pool management all within Rust and can be tweaked via Kconfig.

Some shortcomings:
- This is NOT priority-aware (unlike P4WQ), so in the near-future would like to add that functionality, but want to keep this initial PR short for now.
- This uses an implementation that does not require alloc (since in the Discord it seems @d3zd3z mentioned he would prefer to move away from alloc-only implementations in this repo) but I want to provide an alloc alternative (if the user enables it) that could make it a bit easier to not have to tweak a few Kconfig knobs.

A few other notes:
- Had to tweak queue to provide a `remove` method and had to tweak `lib.rs` to reexport the `zephyr` crate explicitly as its own name to be able to use `zephyr::thread` internally (though a better approach might be to modify the proc macro to not have this be necessary)
- The module is also gated behind `async-drivers` which is kind of confusing since this isn't necessarily only used by drivers, but the `embassy-sync` dep is gated behind `async-drivers` which this uses (so alternatively could just always bring in `embassy-sync`)


d3zd3z (approved):
My only real concern is the behavior when there are more blocks than threads. But I think that can be defined after getting more use out of this.

kurtjd:
> My only real concern is the behavior when there are more blocks than threads. But I think that can be defined after getting more use out of this.

For sure, this was something I was struggling to come up with a great solution to. I had also thought about adding a `try_run` function or something which will just return with an error if a worker thread isn't immediately available to pick up the operation, but figured it would be best to wait to see more use-cases for this. Thanks for the review!

d3zd3z:
I will try using this to make async i2c controller bindings, which can also catch other issues.

zephyr/src/sys/queue.rs
```diff
@@ -73,6 +73,19 @@ impl Queue {
...
+    /// Remove a specific element from the queue, if it is still present.
+    ///
+    /// Returns `true` if `data` was found in the queue and removed, `false` otherwise (for
+    /// example, it was never queued, or has already been received by another thread).
+    ///
+    /// # Safety
+    ///
+    /// `data` must be a pointer previously passed to [`send`](Self::send) on this queue (or null).
+    /// If it is removed, ownership returns to the caller, exactly as with [`recv`](Self::recv).
+    pub unsafe fn remove(&self, data: *mut c_void) -> bool {
```
felipebalbi, line 86: Looks like this could be made fallible:

```rust
pub unsafe fn remove(&self, data: *mut c_void) -> Result<(), Error> {
```

The `bool` could be mapped into a `Result`, that being said, maybe that's better done at a higher layer building a safe API on top of this.

zephyr/src/blocking.rs
```diff
@@ -0,0 +1,405 @@
...
+    BlockingFuture {
+        shared: Shared {
+            op: UnsafeCell::new(Some(op)),
+            result: UnsafeCell::new(None),
+        },
+        slot: None,
+        _pin: PhantomPinned,
+    }
+    .await
+}
```
felipebalbi, line 102: You can save on binary size by changing the return value and patching the fn accordingly:

```rust
pub fn run<F, R>(op: F) -> impl Future<Output = R>
where
    F: FnOnce() -> R + Send + 'static,
    R: Send + 'static,
{
    BlockingFuture {
        shared: Shared {
            op: UnsafeCell::new(Some(op)),
            result: UnsafeCell::new(None),
        },
        slot: None,
        _pin: PhantomPinned,
    }
}
```

zephyr/src/blocking.rs
```diff
@@ -0,0 +1,405 @@
...
+/// Fixed-size storage for internal bookkeeping data.
+///
+/// Because this is stored within a Zephyr queue,
+/// it must be `#[repr(C)]` and the first field is reserved for the kernel's use.
+#[repr(C)]
+struct Slot {
+    /// Reserved for `k_queue`'s linkage while enqueued.
+    _link: UnsafeCell<usize>,
+    /// Serializes access to the future's [`Shared`] state.
+    lock: SpinMutex<()>,
```
felipebalbi, line 129: by `Shared` I suppose you mean the `shared` member below? If that's the case, should this `SpinMutex<UnsafeCell<*mut ()>>`?

felipebalbi (approved):
Looks good, couple minor nits follow.

d3zd3z:
@felipebalbi I think I'm ok merging this, and we can consider the suggestions for subsequent changes, if needed.

zephyr/src/blocking.rs
kurtjd: Hm I think in this case though the return type is no longer `Output = R` but `Output = BlockingFuture<F, R>` right? And I think this would require making BlockingFuture public (and you'd likely want a `#must_use` so the user doesn't forget to await it).

Agree though that it should help decrease binary size slightly.

zephyr/src/sys/queue.rs
kurtjd: Yeah, that is a good point and sounds reasonable, though I agree I think that might belong in a higher-level API if/when we decide to write that (just because all other existing methods here closely match the underlying C API so just wanted to remain consistent).

zephyr/src/blocking.rs
kurtjd: Er maybe not after thinking about it again, need to actually give it a try.

zephyr/src/blocking.rs
kurtjd: So I should probably reword that comment because it's confusing. This lock doesn't JUST guard access to the pointer, it guards several instructions in `run_op`/`Drop` (it's acting similarly to a critical section, but its a bit more performant for this use case).

I could reword that doc string or try to change this a bit such that the SpinMutex is actually encapsulating its entire state.

kurtjd:
> @felipebalbi I think I'm ok merging this, and we can consider the suggestions for subsequent changes, if needed.

Sounds good, yeah happy to address a few things now or wait in follow-up PRs.

zephyr/src/blocking.rs
kurtjd: Yep I was wrong, this works no problem. Can address this now if desired.

felipebalbi:
> @felipebalbi I think I'm ok merging this, and we can consider the suggestions for subsequent changes, if needed.

Works for me.

kurtjd:
Opened a follow up PR to address several review comments: https://github.com/zephyrproject-rtos/zephyr-lang-rust/pull/174

PR #176, opened 2026-08-24
Add AArch64 (arm64) target support

jetpax:
Adds the missing target mapping for AArch64, making Rust available on
arm64 boards.

- `CMakeLists.txt`: map `CONFIG_ARM64` to `aarch64-unknown-none`, or
  `aarch64-unknown-none-softfloat` when `CONFIG_FPU=n`. Zephyr does not
  build arm64 with `-mgeneral-regs-only` and `CPU_CORTEX_A` implies FPU,
  so hard-float is the ABI match for a normal build; the softfloat arm
  keeps a no-FPU board from receiving SIMD/FP register uses its context
  switch never preserves.
- `Kconfig`: `RUST_SUPPORTED` gates on architecture, so the mapping
  alone is never reached — add `ARM64`. The companion
  `!TIMER_READS_ITS_FREQUENCY_AT_RUNTIME` condition is satisfied (the
  ARM generic timer does not select it).
- `etc/platforms.txt`: add `qemu_cortex_a53` so CI can exercise the
  architecture without hardware.

Verified beyond compiling: `samples/philosophers` runs on
`qemu_cortex_a53`. Clippy passes at the project gate (`-D warnings -D
clippy::undocumented_unsafe_blocks`). Notably there was no
`compiler_builtins`-vs-runtime-library link friction on either aarch64
target, so nothing like the riscv workaround was needed.


etc/platforms.txt
```diff
@@ -2,6 +2,7 @@
 -p qemu_cortex_m3
 -p qemu_riscv32
 -p qemu_riscv64
+-p qemu_cortex_a53
```
kurtjd, line 5: I'm not sure if this will actually cause any CI tests to happen on this platform until a sample actually adds `qemu_cortex_a53` to its `platform_allow` list (such as `async_philosophers`). Can wait to see after the first workflow run to double-check before making that change.

You might also need to modify `build.yml` to add `aarch64-zephyr-elf` for the `Setup Zephyr project` step.

CMakeLists.txt
```diff
@@ -49,6 +49,17 @@ function(_rust_map_target)
     else()
       message(FATAL_ERROR "Rust: Unsupported riscv ISA")
     endif()
+  elseif(CONFIG_ARM64)
+    # Zephyr does not build AArch64 with -mgeneral-regs-only, and
+    # CPU_CORTEX_A implies both FPU and FPU_SHARING, so the hard-float
```
kurtjd, line 54: What if the user then manually turns off `FPU_SHARING`? So my understanding is `CPU_CORTEX_A` will imply it by default, but nothing stops the user from overriding that default by disabling it. Should we check for both `CONFIG_FPU` AND `CONFIG_FPU_SHARING`?

dkalowsk (approved):

CMakeLists.txt
```diff
@@ -49,6 +49,17 @@ function(_rust_map_target)
...
+    # target is the ABI match for a normal build.  Boards that turn FPU
+    # off need the softfloat target instead, or Rust would emit SIMD/FP
+    # register uses that the context switch does not preserve.
+    if(CONFIG_FPU)
+      set(RUST_TARGET "aarch64-unknown-none" PARENT_SCOPE)
+    else()
+      set(RUST_TARGET "aarch64-unknown-none-softfloat" PARENT_SCOPE)
+    endif()
```
dkalowsk, line 62: After looking into @kurtjd comment more, I'm not sure this is the right conditional check.  Wouldn't `CONFIG_FP_HARDABI` and `CONFIG_FP_SOFTABI` be better at this point?

PR #177, opened 2026-08-24
fix Kconfig triggered rebuilds and use aligned allocation for Rust

ChidvilasYQC:


kurtjd:
Think it might be preferrable to defer to #176 for enabling ARM64 support (since it more broadly checks `CONFIG_ARM64`) which should cover armv8 and armv9. Though the cargo rerun typo fix and change to allocation behavior could be useful, so could update this PR to just keep those or close this and create another.

Just my recommendation though.

ChidvilasYQC:
@kurtjd thanks for pointing that, I have removed changes related to armv8 from this PR and updated title

eHammarstrom (approved):

dkalowsk (approved):

kurtjd (approved):

PR #179, opened 2026-09-21
zephyr-build: generate all DT properties and fix Parent augment

d3zd3z:
## Summary

- Fix `RawInfo::Parent` in devicetree code generation to check ancestor node status before emitting `get_instance()` calls — prevents compile errors on boards that disable nodes like `pwm-leds` while leaving their children in the tree
- Generate all raw properties from the devicetree as static data accessible via a `Value` enum
- Detect GPIO phandles in raw properties and emit a `Gpio` variant
- Expose more of the GPIO plumbing (`pub(crate)` → `pub`) for applications
- Add runtime `Bytes` devicetree value for byte properties in no_std builds

## Why this is a new PR

This is the same work as #146, unchanged: the five commits here are
byte-identical to the ones approved there by @felipebalbi and @kurtjd.

#146 was targeted at `davidb-ci-davidb-base` rather than `main`, so
merging it put the work on that branch instead of upstream, and GitHub
closed it as merged. Since a merged PR cannot be reopened, this PR
re-proposes the same commits against `main` now that #162 (the CI
trigger those branches were stacked on) has landed.

@felipebalbi @kurtjd — sorry for the churn; a re-approval here would
carry your earlier review across.

## Relationship to other PRs

This is the base of the device-binding series. #148 (PWM LED), #149
(LED strip) and #150 (I2C) are stacked above it and will be retargeted
to `main` as each one lands.

## Testing

These commits are unchanged since the review on #146, and that PR's CI
was green. Since then they have also been carried in the combined branch
used to build the `jolt` keyboard firmware, where the following build and
link cleanly: `samples/hello_world`, `samples/embassy`, `tests/time` and
`tests/drivers/i2c-{controller,target}` on tiny2040,
`tests/drivers/gpio-async` on rpi_pico, and `tests/drivers/led-strip` on
sparkfun_pro_micro_rp2040 — the last of which exercises the `Parent`
augment fix on a board where the `pwm-leds` node is absent.

Not re-verified on this branch alone: the Clippy sweep under Rust 1.98
was run against current `main`, not against these commits in isolation.
CI here covers that, since it builds the branch merged with `main`.


felipebalbi (approved):
Confirmed that the PR is essentially the same.

kurtjd (approved):

dkalowsk (approved):

